PR #7, opened 2020-11-20
producer base setup

shivasishdas:


Makefile
```diff
@@ -20,7 +20,7 @@ PROTO_GIT_URL = $(DRONE_PROTO_GIT_URL)
 endif
 
 # Accept a branch from which we need to checkout proto files. Useful for dev testing.
-PROTO_BRANCH ?= master
+PROTO_BRANCH ?= producer_base
```
Abhishekvrshny, line 23: I dont the variable `PROTO_BRANCH` is being used anywhere. This can be dropped.

grpc_api_configuration.yaml
```diff
@@ -5,3 +5,6 @@ http:
   rules:
     - selector: rzp.common.health.v1.HealthCheckAPI.Check
       get: "/v1/healthcheck"
+    - selector: rzp.metro.producer.v1.ProducerApi.Produce
```
Abhishekvrshny, line 8: what does this configuration do?

service/producer/kafka.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/google/uuid"
+	"github.com/razorpay/metro/internal/config"
+)
+
+type KafkaProducer struct {
+	// will hold the kafka client impl
+}
+
+func NewKakfaProducer(config *config.ConnectionParams) IProducer {
```
Abhishekvrshny, line 12: should there be different implementations of producer for kafka and pulsar? At what level should the segregation of message brokers done?

service/producer/produer.go
```diff
@@ -0,0 +1,27 @@
...
+type IProducer interface {
+	PublishMessage(topic string, message []byte) (string, error)
+}
+
+const (
+	VariantKafka = iota
+	VariantPulsar
+)
+
+func NewProducer(config config.Producer) IProducer {
```
Abhishekvrshny, line 16: Can this be renamed to MessageBroker and be moved to `pkg` package?

service/producer/service.go
```diff
@@ -15,29 +19,40 @@ type Service struct {
...
 
 func (svc *Service) Start() {
-	// Define server handlers
 
 	healthCore, err := health.NewCore(nil)
 	if err != nil {
 		panic(err)
 	}
 
+	producer := NewProducer(boot.Config.Producer)
```
Abhishekvrshny, line 36: Make all singleton initializations in boot and pass it here. Essentially follow dependency injection pattern

grpc_api_configuration.yaml
shivasishdas: Maps the REST endpoints to the RPC

Makefile
shivasishdas: Removed.

grpc_api_configuration.yaml
```diff
@@ -5,3 +5,6 @@ http:
...
+      post: "/v1/produce"
```
amitmahbubani, line 9: Are we going for RESTful APIs? If yes, this is not RESTful, the API should be on top of a resource which would be a noun, produce is an action/verb. `POST /v1/messages` for example fits.

service/producer/service.go
shivasishdas: Yup this is a singleton initialization.

grpc_api_configuration.yaml
Abhishekvrshny: Agree. Also, this file is not required now. Use the right annotations in metro-proto itself. Please use this as reference https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto and we should be able to copy the same terminology

amitmahbubani: 👍

deployment/dev/docker-compose-kakfa.yml
```diff
@@ -0,0 +1,148 @@
...
+      KSQL_BOOTSTRAP_SERVERS: kafka1:9092
+      KSQL_LISTENERS: http://0.0.0.0:8088
+      KSQL_KSQL_SERVICE_ID: kafka_workshop
+      KSQL_CUB_KAFKA_TIMEOUT: 300
+      KSQL_KSQL_SCHEMA_REGISTRY_URL: http://schemaregistry:8081
+      KSQL_KSQL_COMMIT_INTERVAL_MS: 2000
+      KSQL_KSQL_CACHE_MAX_BYTES_BUFFERING: 10000000
+      KSQL_KSQL_STREAMS_AUTO_OFFSET_RESET: earliest
+
+  database:
```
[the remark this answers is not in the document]
shivasishdas: was trying out all the Kafka features and checking its internals

Abhishekvrshny (approved):

grpc_api_configuration.yaml
varunachar: Noob here about RPCs in general, but from my understanding, an RPC call is like a Command, so why does it have to be a RESTful API? Isn't `/publish` a better endpoint to have?

And similarly, why is it `ProducerApi.Produce` and not `ProducerApi.Publish`?

varunachar: The reference shared by Abhishek above also is going for a Command paradigm. If I'm reading it correctly, but aren't the two endpoints `.../acknowledge` and `.../pull`?

image

amitmahbubani: > RPC call is like a Command, so why does it have to be a RESTful API?

You're right, it doesn't. I was saying "if" we're doing REST-ful, let's stick to it's guidelines. If not, RPC-like APIs are good, so `/publish` or `/publishMessage` will fit well.


> The reference shared by Abhishek above also is going for a Command paradigm. If I'm reading it correctly, but aren't the two endpoints .../acknowledge and .../pull?

Google OTOH has a custom spec for representing actions in a REST API, [see here](https://cloud.google.com/apis/design/custom_methods). So you're looking at REST APIs with a command attached, above the default GET/POST/... standard actions. The endpoint is [/v1/subscription:acknowledge](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/acknowledge).

PR #8, opened 2020-11-25
integrate with buf

Abhishekvrshny:
- [X] replaces `prototool` with `buf`
- [X] `grpc-gateway` works with annotations in `.proto` files. This is different from how `prototool` used to work with `grpc-gateway`
- [X] Exposes openAPI (swagger) endpoint on port 3000

TODO:
- [ ] Make swagger listen on same port as `grpc-gateway` to enable `try it out` functionality

Note:
1. Reference repo: https://github.com/johanbrandhorst/grpc-gateway-boilerplate 
2. `buf` is maintained by the maintainers of grpc itself. johanbrandhorst also confirmed that its ready for production use. https://github.com/johanbrandhorst
3. `third_party` directory is required and contains copied proto definitions. `buf` is building a schema registry for protobuf to solve for this.


Makefile
```diff
@@ -1,4 +1,14 @@
 # Dir where build binaries are generated. The dir should be gitignored
+
+## Set defaults
+export GO111MODULE := on
+export PATH := $(GOBIN):$(PATH)
```
sunnyagg, line 5: this is required in general for local go setup, do we need it as part of makefile, we can add this in readme ?

sunnyagg (approved):

Makefile
Abhishekvrshny: `Makefile` is used in CI runs as well.

PR #10, opened 2020-11-26
[consul] adds consul client and helper methods

sunnyagg:
- Adds consul client
- Few helper methods which can be used for leader election

This might be non working code. will setup consumers binary first and will be fixed when consumer pod uses this


Abhishekvrshny (approved):

pkg/registry/consul.go
```diff
@@ -0,0 +1,115 @@
+package registry
+
+import (
+	"github.com/hashicorp/consul/api"
```
Abhishekvrshny, line 4: you can run `go imports` to fix linting and alignments.

PR #14, opened 2020-11-30
Push/Pull Consumer

sunnyagg:
- Refactor code to support multiple services
- Adds support for service level configurations


cmd/service/main.go
```diff
@@ -2,39 +2,63 @@ package main
...
 import (
 	"context"
+	"flag"
 	"log"
 	"os"
 	"os/signal"
 	"syscall"
 
+	"github.com/razorpay/metro/cmd/service/metro"
+
```
Abhishekvrshny, line 12: nit: extra space

cmd/service/metro/metro.go
```diff
@@ -0,0 +1,81 @@
...
+func isValidService(service string) bool {
+	for _, s := range validServices {
+		if s == service {
+			return true
+		}
+	}
+	return false
+}
+
+type Server struct {
```
Abhishekvrshny, line 33: a better name for this struct? For eg, a push-consumer is not really a server.

Abhishekvrshny (approved):

config/default.toml
```diff
@@ -55,3 +50,14 @@
     dsn                   = "dummy"
     enabled               = false
     logLevel              = "Error"
+
+[services]
```
Abhishekvrshny, line 54: TODO: Think if this nesting is required or not.

Abhishekvrshny (approved):

cmd/service/main.go
sunnyagg: it looks aligned to me :(

config/default.toml
sunnyagg: i have 2 things in mind, i am confused on 2 structures (will do in other PR to keep this one short), also need to merge all dev, docker and normal one into one config, all this will be handlled in config related changes PR

current thoughts
1.
```
[producer]
a=b
a1=b1
[push-consumer]
a=b
a2=b2
[pull-consumer]
a=b
a3=b3
```
this will completely separate out the configs for each service, we can have separate files itself then, there will be duplicates of settings here but easy to make changes with respect to 1 service without impacting others

2.
[common]
a=b
[producer]
a1=b1
[push-consumer]
a2=b2
[pull-consumer]
a3=b3

merging configs can be pain here but still doable

PR #16, opened 2020-12-08
Makefile enhancement and packages cleanup

Abhishekvrshny:
- [X] Adds `goimports` target in Makefile
- [X] Cleanup of the Makefile
- [X] Removes unused packages like worker, job, spine etc. Can be added later if required

TODO:
- [ ] Enable limiting and goimports check in CI.


config/default.toml
```diff
@@ -24,24 +24,6 @@
     localAgentHostPort    = "localhost:6831"
     disabled              = true
 
-[queue]
```
sunnyagg, line 27: can we remove mysql db as well ?

Abhishekvrshny: @sunnyagg done

sunnyagg (approved):

PR #20, opened 2020-12-10
[Consul] Adds consul config support

sunnyagg:
- Adds config for consul
- Setup Consul on app startup


Abhishekvrshny (approved):

config/default.toml
```diff
@@ -20,12 +20,18 @@
         [producer.broker.BrokerConfig]
             [producer.broker.BrokerConfig.Producer]
                 Brokers       = ["localhost:9021"]
+    [producer.registry]
+        driver                    = "consul"
+        name                      = "metro"
```
Abhishekvrshny, line 25: what is the significance of name here?

PR #22, opened 2020-12-12
remove third_party proto defs

Abhishekvrshny:
- [X] third_party proto definitions are moved to metro-proto repo, hence removing it from this repo.
- [X] There is a slight change in rpc import paths.


shivasishdas (approved):
LGTM

PR #24, opened 2020-12-15
config refactor and openapi-server component

Abhishekvrshny:
- [X] Move component config to their respective packages, making it more go-like.
- [X] Remove component config map as it introduces the limitation of all component configs to be alike, which may not be the case.
- [X] Move common config to top-level. The idea here is that any initialization which requires common config should happen before the component is initialized and component level initialization can happen in the component itself with its own config. 
- [X] These changes also prevent repetitive configs across components.
- [X] Also introduces openapi-server as an independent component.
- [X] Executing APIs through swagger is also functional now.
- [X] Some refactoring on `internal/server`


config/dev_docker.toml
```diff
@@ -1,101 +1,71 @@
+[app]
+    appEnv                = "default"
+    serviceName           = "metro-producer"
```
sunnyagg, line 3: this needs to be common now `metro`

sunnyagg (approved):

config/dev_docker.toml
Abhishekvrshny: done

PR #25, opened 2020-12-21
implements creating a project entity in registry

Abhishekvrshny:
Also introduces interfaces and base structs for models, repos with
some refactoring


internal/common/repo.go
```diff
@@ -0,0 +1,25 @@
...
+	Registry registry.Registry
+}
+
+// Create puts the model in the registry
+func (r Repo) Create(m Model) error {
+	b, err := json.Marshal(m)
+	if err != nil {
+		return err
+	}
+	return r.Registry.Put(basePrefix+m.Prefix()+m.Key(), b)
```
sunnyagg, line 24: should this be exposed by a method in model ?
getKey(){
return basePrefix+m.Prefix()+m.Key()
}

sunnyagg (commented):
add tests :)

internal/common/repo.go
Abhishekvrshny: I think prefix and key can come from Key() method, but base prefix can be in base class itself?

Abhishekvrshny: on the other hand, for list functionality, I just need basePrefix+m.Prefix(), so the current scheme works there. Thoughts?

sunnyagg (approved):

PR #26, opened 2020-12-21
Adds leaderelection package for push-consumer

sunnyagg:
- Adding leader election using consul for push-consumer


pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,203 @@
+package leaderelection
+
+import (
+	"context"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/runtime"
+
```
Abhishekvrshny, line 8: nit: remove the extra line breaks and run `make goimports`

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,203 @@
...
+func (le *LeaderElector) Run(ctx context.Context) {
+	defer runtime.HandleCrash()
+	defer func() {
+		le.config.Callbacks.OnStoppedLeading()
+	}()
+
+	if !le.acquire(ctx) {
+		return // ctx signalled done
+	}
+	ctx, cancel := context.WithCancel(ctx)
```
Abhishekvrshny, line 62: what happens if `ctx` passed is already a context with cancel?

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,203 @@
...
+	if le.nodeID != "" {
+		// if node is not registred anymore, we will need register it again
+		// it can happen that ttl associated with registration had expired
+		if ok := le.registry.IsRegistered(le.nodeID); !ok {
+			logger.Log.Info("node is unregisterd")
+			le.nodeID = ""
+		} else {
+			err = le.registry.Renew(le.nodeID)
+			if err != nil {
+				logger.Log.Infof("error while renewing registration: %v", err.Error())
```
Abhishekvrshny, line 117: nit: Infow() can be used for structured logging with which takes (....,key, value, key, value) as args

Abhishekvrshny: And meta can be attached to all logs this way

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,203 @@
...
+			}
+		}
+	}
+
+	// if not registered, register it
+	if le.nodeID == "" {
+		le.nodeID, err = le.registry.Register(le.config.Name, le.config.LeaseDuration)
+		if err != nil {
+			logger.Log.Errorf("failed to register node with registry: %v", err)
+			return false
```
Abhishekvrshny, line 128: should bool be returned only when leader is acquired or not. In case of error in registering or renewing lease, shouldn't we return error and let it be handled at upper layers where the consumer can even shutdown?

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,203 @@
...
+		le.config.Callbacks.OnStoppedLeading()
+	}()
+
+	if !le.acquire(ctx) {
+		return // ctx signalled done
+	}
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	go le.config.Callbacks.OnStartedLeading(ctx)
+	le.renew(ctx)
```
Abhishekvrshny, line 65: nit: leaderelection and registry both have `renew` functions. Have different names?

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,203 @@
...
+	defer func() {
+		le.config.Callbacks.OnStoppedLeading()
+	}()
+
+	if !le.acquire(ctx) {
+		return // ctx signalled done
+	}
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	go le.config.Callbacks.OnStartedLeading(ctx)
```
Abhishekvrshny, line 64: how do we capture error in `OnStartedLeading ` ?

pkg/leaderelection/leaderelection.go
Abhishekvrshny: if this function returns, it would silently terminate the goroutine. The application should know that this has returned and probably shutdown the app?

service/push-consumer/service.go
```diff
@@ -13,13 +20,15 @@ type Service struct {
...
 	registry registry.Registry
+	NodeID   string
 }
 
 // NewService creates an instance of new push consumer service
 func NewService(ctx context.Context, config *Config) *Service {
 	return &Service{
 		ctx:    ctx,
 		config: config,
+		NodeID: uuid.New().String(),
```
Abhishekvrshny, line 31: Should we take the k8s stateful set as the node id?

service/push-consumer/service.go
```diff
@@ -2,6 +2,13 @@ package pushconsumer
 
 import (
 	"context"
+	"time"
+
+	"github.com/razorpay/metro/pkg/logger"
+
```
Abhishekvrshny, line 8: nit: goimports

pkg/registry/registry.go
```diff
@@ -1,20 +1,28 @@
...
+	// Renew a regestration using registration_id
+	Renew(string) error
+
 	// Acquire a lock for a registration_id on a given key and value pair
-	Acquire(string, string, string) (bool, error)
+	Acquire(string, string, string) bool
 
 	// Release a lock for a restration_id on a given key and value pair
-	Release(string, string, string) (bool, error)
+	Release(string, string, string) bool
```
Abhishekvrshny, line 25: error may still need to be handled differently?

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// registry being used for leader election
+	registry registry.Registry
+}
+
+// Run starts the leader election loop. Run will not return
+// before leader election loop is stopped by ctx or it has
+// stopped holding the leader lease
+func (le *LeaderElector) Run(ctx context.Context) {
+	defer runtime.HandleCrash()
```
Abhishekvrshny, line 54: HandleCrash() does `recover()` which might be riskier, go doesn't encourage recovering. Lets remove this for now

pkg/leaderelection/leaderelection.go
sunnyagg: had run it, changing it manually

pkg/leaderelection/leaderelection.go
sunnyagg: will remove

pkg/leaderelection/leaderelection.go
sunnyagg: ctx works on parent-child concept and we should propagate the cancellation, i guess that is what  is not handled properly

pkg/leaderelection/leaderelection.go
sunnyagg: Good Question, Will add a todo and get back

pkg/leaderelection/leaderelection.go
sunnyagg: adding a TODO for this too, had thought of this, didn't spent time on solving

pkg/leaderelection/leaderelection.go
sunnyagg: There are 2 infinite loops running outside
1. Acquiring - This cares only about acq and retry on all failures
2. Renew - This cares only about renew success and retry for all failures

hence kept it same way, otherwise upper layer will need to handle both err and bool but action will be same to retry

pkg/registry/registry.go
sunnyagg: currently it was getting handled same way so removed, can add back if we need different handling

service/push-consumer/service.go
sunnyagg: Yes, i remember that discussion, we should do that, just keeping some random id for start, will add todo

pkg/leaderelection/leaderelection.go
Abhishekvrshny: But infinite retry is also problematic. Lets discuss this

Abhishekvrshny (approved):

PR #32, opened 2020-12-27
various changes related to tests, dummy server impl etc

Abhishekvrshny:
- [X] introduces google pub/sub emulator for running compatibility tests of metro's APIs against that of pub/sub
- [X] support for running integration tests and compatibility tests using build tags in go, see `make help`
- [X] dummy implementation for various topic and subscription APIs for an MVP publish/consume flow. Related PR: https://github.com/razorpay/metro-proto/pull/11
- [X] enhancing project create API with validations
- [X] split github workflow into multiple jobs (unit-tests, build and integration tests)
- [X] misc cleanup

TODO
- [X] ~compatibility tests need to be incorporated in github actions and docker-compose locally. They are commented out here for now, see https://github.com/razorpay/metro/pull/32/files#diff-05d0c5ed3704411b87853f12c9e2353fc4b0b56e500c573e59ec1ce8d51f5993R18~ (done)
- [X] ~update readme~ (done)


Makefile
```diff
@@ -140,6 +152,22 @@ mock-gen-clean:
 docs-uml:
 	@go-plantuml generate --recursive --directories cmd --directories internal --directories pkg --out $(DOCS_DIR)/$(UML_OUT_FILE)
 
+.PHONY: test-integration-ci ## run integration tests on ci (github actions)
+test-integration-ci:
+	@METRO_TEST_HOST=metro-producer go test ./... -tags=integration,musl
```
sunnyagg, line 157: does this mean these tests are being run only for producer component ?

tests/integration/sample_test.go
```diff
@@ -0,0 +1,18 @@
...
+	"fmt"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_Topic_CreateTopic(t *testing.T) {
+	url := fmt.Sprintf("http://%s:8082/v1/healthcheck", os.Getenv("METRO_TEST_HOST"))
```
sunnyagg, line 15: if all components are running, how do we know which component is running on which port on local, should we read it from config somehow

sunnyagg (approved):

Makefile
Abhishekvrshny: Integration tests run against API endpoints which is being exposed by producer today. I am thinking we can rename it as `web` and have both producer and pull consumer APIs be part of it?

tests/integration/sample_test.go
Abhishekvrshny: we'll have to parse config files for this in the test case. I guess we can live by convention?

PR #34, opened 2020-12-29
leaderelection handling for sigterm and code cleanup

sunnyagg:
- Sigterm handling for leaderelection
- Code Cleanup and error handling for go routines
- Cancel context handling and cancellation propogation

### DIFF
--- a/cmd/service/metro/metro.go
+++ b/cmd/service/metro/metro.go
@@ -80,22 +80,28 @@ func Run(ctx context.Context) {
 		}
 	}()
 
-	err := component.Start()
-	if err != nil {
-		logger.Ctx(ctx).Fatalw("error in starting component", "msg", err.Error())
-		return
-	}
+	// Handle error from component start
+	errCh := make(chan error)
+
+	// start the component in a go routine, Start should be implemented as a blocking call
+	go func() {
+		errCh <- component.Start()
+	}()
 
 	// Handle SIGINT & SIGTERM - Shutdown gracefully
-	c := make(chan os.Signal, 1)
-	signal.Notify(c, syscall.SIGTERM, syscall.SIGINT)
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
 
+	select {
 	// Block until SIGINT/SIGTERM signal is received
-	sig := <-c
-	logger.Ctx(ctx).Infow("received a signal, stopping metro", "signal", sig)
+	case sig := <-sigCh:
+		logger.Ctx(ctx).Infow("received a signal, stopping metro", "signal", sig)
+	case err := <-errCh:
+		logger.Ctx(ctx).Fatalw("component exited with error", "msg", err.Error())
+	}
 
-	// call stop component, it should internally clean up all running go routines
-	err = component.Stop()
+	// call stop component, it should internally clean up for all running go routines
+	err := component.Stop()
 	if err != nil {
 		logger.Ctx(ctx).Fatalw("error in stopping metro")
 	}
--- a/cmd/service/metro/metro.go
+++ b/cmd/service/metro/metro.go
@@ -80,24 +80,22 @@ func Run(ctx context.Context) {
 		}
 	}()
 
-	errChan := component.Start()
+	err := component.Start()
+	if err != nil {
+		logger.Ctx(ctx).Fatalw("error in starting component", "msg", err.Error())
+		return
+	}
 
 	// Handle SIGINT & SIGTERM - Shutdown gracefully
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, syscall.SIGTERM, syscall.SIGINT)
 
-	// Block until signal is received or error is received in starting the component.
-	select {
-	case <-c:
-		logger.Ctx(ctx).Infow("received signal")
-	case err := <-errChan:
-		logger.Ctx(ctx).Fatalw("error in starting component", "msg", err.Error())
-	}
+	// Block until SIGINT/SIGTERM signal is received
+	sig := <-c
+	logger.Ctx(ctx).Infow("received a signal, stopping metro", "signal", sig)
 
-	logger.Ctx(ctx).Infow("stopping metro")
-
-	// stop component
-	err := component.Stop()
+	// call stop component, it should internally clean up all running go routines
+	err = component.Stop()
 	if err != nil {
 		logger.Ctx(ctx).Fatalw("error in stopping metro")
 	}
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -1,9 +1,10 @@
 package server
 
 import (
-	"context"
 	"net"
 	"net/http"
+
+	"golang.org/x/sync/errgroup"
 
 	grpcmiddleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpcctxtags "github.com/grpc-ecosystem/go-grpc-middleware/tags"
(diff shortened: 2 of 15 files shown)


cmd/service/metro/metro.go
```diff
@@ -80,24 +80,22 @@ func Run(ctx context.Context) {
 		}
 	}()
 
-	errChan := component.Start()
+	err := component.Start()
```
Abhishekvrshny, line 83: The component here can be encapsulated into a struct along with ctx. This would avoid passing the same ctx twice i.e in Init() and Run()

internal/server/server.go
```diff
@@ -18,12 +19,13 @@ type registerGrpcHandlers func(server *grpc.Server) error
...
 
 // StartGRPCServer with handlers and interceptors
-func StartGRPCServer(errChan chan<- error, address string, registerGrpcHandlers registerGrpcHandlers, interceptors ...grpc.UnaryServerInterceptor) (*grpc.Server, error) {
+func StartGRPCServer(ctx context.Context, address string, registerGrpcHandlers registerGrpcHandlers, interceptors ...grpc.UnaryServerInterceptor) (*grpc.Server, error) {
 	grpcServer, err := newGrpcServer(registerGrpcHandlers, interceptors...)
 	if err != nil {
 		return nil, err
 	}
 	// Start gRPC server.
+	var errChan chan<- error
```
Abhishekvrshny, line 28: what is the use of this `errChan` if its not being passed as function argument?

internal/server/server.go
```diff
@@ -18,12 +19,13 @@ type registerGrpcHandlers func(server *grpc.Server) error
 type registerHTTPHandlers func(mux *runtime.ServeMux) error
 
 // StartGRPCServer with handlers and interceptors
-func StartGRPCServer(errChan chan<- error, address string, registerGrpcHandlers registerGrpcHandlers, interceptors ...grpc.UnaryServerInterceptor) (*grpc.Server, error) {
+func StartGRPCServer(ctx context.Context, address string, registerGrpcHandlers registerGrpcHandlers, interceptors ...grpc.UnaryServerInterceptor) (*grpc.Server, error) {
```
Abhishekvrshny, line 22: where is this ctx used?

internal/server/server.go
```diff
@@ -39,12 +41,13 @@ func StartGRPCServer(errChan chan<- error, address string, registerGrpcHandlers
 }
 
 // StartHTTPServer with handlers
-func StartHTTPServer(errChan chan<- error, address string, registerHTTPHandlers registerHTTPHandlers) (*http.Server, error) {
+func StartHTTPServer(ctx context.Context, address string, registerHTTPHandlers registerHTTPHandlers) (*http.Server, error) {
```
Abhishekvrshny, line 44: same

internal/server/server.go
```diff
@@ -39,12 +41,13 @@ func StartGRPCServer(errChan chan<- error, address string, registerGrpcHandlers
...
 	httpServer, err := newHTTPServer(registerHTTPHandlers)
 	if err != nil {
 		return nil, err
 	}
 	// Start HTTP server for gRPC gateway.
+	var errChan chan<- error
```
Abhishekvrshny, line 50: same

pkg/registry/consul.go
```diff
@@ -43,8 +43,9 @@ func NewConsulClient(config *ConsulConfig) (IRegistry, error) {
 // Register is used for node registration which creates a session to consul
 func (c *ConsulClient) Register(name string, ttl time.Duration) (string, error) {
 	sessionID, _, err := c.client.Session().Create(&api.SessionEntry{
-		Name: name,
-		TTL:  ttl.String(),
+		Name:      name,
+		TTL:       ttl.String(),
+		LockDelay: 2,
```
Abhishekvrshny, line 48: from the documentation https://www.consul.io/docs/dynamic-app-config/sessions, it says
```
While the default is to use a 15 second delay, clients are able to disable this mechanism by providing a zero delay value.
```
Is 2s a good value for us?

service/push-consumer/service.go
```diff
@@ -17,7 +18,10 @@ type Service struct {
 	health   *health.Core
 	config   *Config
 	registry registry.IRegistry
+	le       *leaderelection.LeaderElector
```
Abhishekvrshny, line 21: nit: `leaderelection.LeaderElector` has repetitive words and go discourages this, how about leaderelection.Controller or leaderelection.Commisioner or something else?

pkg/leaderelection/leaderelection.go
```diff
@@ -2,18 +2,14 @@ package leaderelection
...
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-const (
-	// JitterFactor used in wait utils
-	JitterFactor = 1.2
-)
-
 // NewLeaderElector creates a LeaderElector from a LeaderElection Config
 func NewLeaderElector(c Config, registry registry.IRegistry) (*LeaderElector, error) {
```
Abhishekvrshny, line 14: as per go convention packageName.New() is preffered over packageName.NewPackageName()

pkg/leaderelection/leaderelection.go
```diff
@@ -47,18 +43,42 @@ type LeaderElector struct {
...
+
+			if !acquired {
+				logger.Ctx(retryCtx).Infow("failed to acquire lease, retrying...")
+				return
+			}
+
+			// if succeeded, we can break the inner loop by cancelling context and renew in outer loop
+			logger.Ctx(ctx).Info("successfully acquired lease")
+			retryCancel()
+		}, le.config.RetryPeriod, retryCtx.Done())
```
Abhishekvrshny, line 70: jitter was good. why did we remove it?

pkg/leaderelection/leaderelection.go
```diff
@@ -47,18 +43,42 @@ type LeaderElector struct {
...
+			acquired = le.tryAcquireOrRenew(retryCtx)
+
+			if !acquired {
+				logger.Ctx(retryCtx).Infow("failed to acquire lease, retrying...")
+				return
+			}
+
+			// if succeeded, we can break the inner loop by cancelling context and renew in outer loop
+			logger.Ctx(ctx).Info("successfully acquired lease")
+			retryCancel()
```
Abhishekvrshny, line 69: This cancel doesn't have any implication right now, correct?

pkg/leaderelection/leaderelection.go
```diff
@@ -47,18 +43,42 @@ type LeaderElector struct {
...
+		// OnStartedLeading
+		le.leaderID = le.nodeID
+		go le.config.Callbacks.OnStartedLeading(leadCtx)
+
+	}, le.config.RenewDeadline, ctx.Done())
+
+	// context returned done, release the lease
+	logger.Ctx(ctx).Info("context done, releasing lease")
+	leadCancel()
+	le.release(ctx)
```
Abhishekvrshny, line 81: would need to wait for `OnStartedLeading` to successfully terminate before returning from this function

pkg/leaderelection/leaderelection.go
sunnyagg: this cancel causes the context done channel to be closed, hence it causes exit from the wait loop

pkg/leaderelection/leaderelection.go
sunnyagg: wasn't sure what its actually doing, it was used at only one place (acquire) but not in renew, let me check again

pkg/leaderelection/leaderelection.go
sunnyagg: release calls onStopLeading callback, that will ensure that we no longer perform leader tasks

pkg/registry/consul.go
sunnyagg: currently we have 5s retry period for reattempting the lock acquire so anything below 5 should be fine

service/push-consumer/service.go
sunnyagg: renaming to Candidate

cmd/service/metro/metro.go
sunnyagg: since errCh was removed from interface, because of that change this code was breaking, this was temp fix to let the build pass and fix rest things, updated

sunnyagg: using same ctx in multiple func should be fine, will think if we need a struct here

Abhishekvrshny (approved):

PR #36, opened 2021-01-04
implements topic create API

Abhishekvrshny:
- [X] also introduces a custom error package

TODO
- [ ] integrate with messagebroker when its APIs are done.
- [ ] integration and compatibility tests


internal/project/validation.go
```diff
@@ -1,18 +1,47 @@
...
+	m.Labels = proto.GetLabels()
+	return m
+}
+
+func isValidProjectID(ctx context.Context, projectID string) bool {
+	if strings.HasSuffix(projectID, "-") {
+		logger.Ctx(ctx).Error("projectID cannot end with a trailing hyphen")
+		return false
+	}
+	r, err := regexp.Compile("([a-z][a-z0-9-]{5,29})$")
```
shivasishdas, line 41:
```suggestion
	var idRegex, _ := regexp.Compile("([a-z][a-z0-9-]{5,29})$")
```
We can export this to a var outside the function and ignore the error since this would be a one-time expr setup. Will avoid recompilation on every function call.

internal/topic/core.go
```diff
@@ -0,0 +1,57 @@
...
+	return &Core{repo, projectCore}
+}
+
+// CreateTopic implements topic creation
+func (c *Core) CreateTopic(ctx context.Context, m *Model) error {
+	if ok, err := c.projectCore.ExistsWithID(ctx, m.ExtractedProjectID); !ok {
+		if err != nil {
+			return err
+		}
+		merror.Newf(merror.NotFound, "project not found")
```
shivasishdas, line 36: return?

internal/topic/validation.go
```diff
@@ -0,0 +1,51 @@
...
+	m.Labels = req.GetLabels()
+	m.ExtractedProjectID = p
+	m.ExtractedTopicName = t
+	return m, nil
+}
+
+func extractTopicMetaAndValidate(ctx context.Context, name string) (projectID string, topicName string, err error) {
+	// https://github.com/googleapis/googleapis/blob/69697504d9eba1d064820c3085b4750767be6d08/google/pubsub/v1/pubsub.proto#L170
+	// Note: check for project ID would happen while creating the project, hence not enforcing it here
+	r, err := regexp.Compile("projects/(.*)/topics/([A-Za-z][A-Za-z0-9-_.~+%]{2,254})$")
```
shivasishdas, line 32: extract out to avoid recompilation?

internal/merror/codes.go
```diff
@@ -0,0 +1,28 @@
...
+	AlreadyExists
+	// NotFound ...
+	NotFound
+	// Unknown ...
+	Unknown
+)
+
+var internalToGRPCCodeMapping map[Code]codes.Code
+
+func init() {
```
shivasishdas, line 21: this is auto-called on boot right?

internal/merror/merror.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"fmt"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// MError is metro error
+type MError struct {
```
shivasishdas, line 11: should we add an optional metadata map as well? Sometimes adding too much context in the error message itself will make it verbose. All such extra (k,v) can be parked inside the metadata map to enrich the error.
```
{
	"code": "TopicCreationFailed",
	"message": "failed to created topic",
	"metadata": {
		"topic": "payments.auth",
		"broker": "kafka",
		"some-use-full-key-in-future": "value"
	}
}
```
Thoughts?

internal/project/core.go
```diff
@@ -23,12 +28,28 @@ func NewCore(repo IRepo) *Core {
...
 	return c.repo.Create(ctx, m)
 }
+
+// Exists to check if the project exists with fully qualified consul key
+func (c *Core) Exists(ctx context.Context, key string) (bool, error) {
+	logger.Ctx(ctx).Infow("exists query on project", "key", key)
+	ok, err := c.repo.Exists(ctx, key)
+	if err != nil {
+		logger.Ctx(ctx).Errorw("error in executing exists", "msg", err.Error())
+		return false, err
```
shivasishdas, line 47: this should be wrapped in `merror.Newf()`?

shivasishdas (changes requested):

internal/project/validation.go
```diff
@@ -1,18 +1,52 @@
...
+func fromProto(proto *metrov1.Project) *Model {
+	m := &Model{}
+	m.Name = proto.GetName()
+	m.ProjectID = proto.GetProjectId()
+	m.Labels = proto.GetLabels()
+	return m
+}
+
+func isValidProjectID(ctx context.Context, projectID string) bool {
+	if strings.HasSuffix(projectID, "-") {
```
sunnyagg, line 47: can this also be validated as part of regex ?

internal/topic/model.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// Key returns the key for storing in the registry
+func (m *Model) Key() string {
+	return m.Prefix() + m.ExtractedTopicName
+}
+
+// Prefix returns the key prefix
+func (m *Model) Prefix() string {
+	return common.BasePrefix + project.Prefix + m.ExtractedProjectID + "/" + Prefix
```
sunnyagg, line 29: probably a helper methods per package

sunnyagg (approved):

internal/project/validation.go
Abhishekvrshny: lookarounds in regex are not supported in golang, hence did it this way. There might be some other way though, didnt spend too much time in that

PR #38, opened 2021-01-05
fix broker configs

shivasishdas:


service/producer/service.go
```diff
@@ -49,7 +49,7 @@ func (svc *Service) Start() error {
 	mb, err := messagebroker.NewProducerClient(context.Background(),
 		messagebroker.Kafka,
 		&svc.config.Broker.BrokerConfig,
-		&messagebroker.ProducerClientOptions{},
+		&messagebroker.ProducerClientOptions{Topic: "dummy-topic"},
```
Abhishekvrshny, line 52: is topic requires here?

Abhishekvrshny (approved):

PR #39, opened 2021-01-05
consul watch implementation

sunnyagg:
- Implements consul watch functionality
- Handles graceful shutdown of go routines for push-consumer

### DIFF
--- a/pkg/leaderelection/candidate.go
+++ b/pkg/leaderelection/candidate.go
@@ -49,6 +49,7 @@ type Candidate struct {
 // stopped holding the leader lease
 func (c *Candidate) Run(ctx context.Context) error {
 	leadCtx, leadCancel := context.WithCancel(ctx)
+	defer leadCancel()
 
 	logger.Ctx(ctx).Info("attempting to acquire leader lease")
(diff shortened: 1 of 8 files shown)


Abhishekvrshny (approved):

pkg/leaderelection/candidate.go
```diff
@@ -80,9 +81,7 @@ func (c *Candidate) Run(ctx context.Context) error {
 
 	// context returned done, release the lease
 	logger.Ctx(ctx).Info("context done, releasing lease")
-	leadCancel()
```
sunnyagg, line 83: this is not required since parent context will cancel it automatically

pkg/registry/watch.go
```diff
@@ -0,0 +1,12 @@
+package registry
+
+// HandlerFunc provides interface for the watch handler func
+// which is implemented by watch subscriber
+type HandlerFunc func([]Pair)
+
+// WatchHandler struct provides watch capability for registry
+type WatchHandler struct {
```
sunnyagg, line 8: this is WIP, this will implement the watch cancelation

PR #42, opened 2021-01-07
handles watch termination

sunnyagg:
- Handles watch termination with registry for graceful shutdown
- Improved test coverage for leaderelection package

### DIFF
--- a/pkg/registry/consul.go
+++ b/pkg/registry/consul.go
@@ -11,10 +11,8 @@ import (
 
 // ConsulClient ...
 type ConsulClient struct {
-	ctx      context.Context
-	client   *api.Client
-	plans    []*watch.Plan
-	watchers []*ConsulWatchHandler
+	ctx    context.Context
+	client *api.Client
 }
 
 // ConsulConfig for ConsulClient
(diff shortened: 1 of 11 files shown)


pkg/registry/consulwatch.go
```diff
@@ -2,22 +2,27 @@ package registry
...
+		plan:   plan,
+		client: client,
 	}
 }
 
 // Handler implements the consul watch handler method and invokes the requester handler
-func (cw *ConsulWatchHandler) Handler(index uint64, result interface{}) {
+func (cwh *ConsulWatcher) handler(index uint64, result interface{}) {
 	pairs, ok := result.(api.KVPairs)
 	if !ok {
```
Abhishekvrshny, line 27: log error, or set error somewhere so that it can be logged outside?

pkg/registry/consul.go
```diff
@@ -110,27 +108,20 @@ func (c *ConsulClient) Release(sessionID string, key string, value string) bool
...
+		"prefix": wh.WatchPath,
 	})
 
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	wh := NewConsulWatchHandler(handler)
+	cwh := NewConsulWatcher(wh, plan, c.client)
```
Abhishekvrshny, line 122: can IWatcher interface be injected in this func. Would help with unit testing

service/push-consumer/service.go
```diff
@@ -86,14 +85,39 @@ func (c *Service) Start() error {
...
+		var err error
+		select {
+		case <-gctx.Done():
+			err = gctx.Err()
+		case <-c.stopCh:
+			err = fmt.Errorf("signal received, stopping push-consumer")
+		}
+
+		if watcher != nil {
+			watcher.StopWatch()
```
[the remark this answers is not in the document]
sunnyagg: This is just a signal, so this signal will cause the actual watch to terminate which is a blocking watch. we are waiting for Actual watch termination

pkg/registry/consul.go
sunnyagg: This func is being used to create watcher, method can be renamed but not sure if interface injection is possible, we will need to create interface in registry

pkg/registry/consulwatch.go
sunnyagg: will discuss

sunnyagg: will fix all error handling in next PR, will merge this

Abhishekvrshny (approved):

PR #44, opened 2021-01-10
adds more apis and compatibility tests

Abhishekvrshny:


internal/project/validation.go
```diff
@@ -21,8 +21,8 @@ func init() {
 }
 
 // GetValidatedModel validates an incoming proto request and returns a project model
-func GetValidatedModel(ctx context.Context, req *metrov1.Project) (*Model, error) {
-	if req.Name == "" {
+func GetValidatedModel(ctx context.Context, req *metrov1.Project, allowEmptyName bool) (*Model, error) {
```
sunnyagg, line 24: this seems hackish, why do we need project with empty name

Abhishekvrshny: delete call takes projectID in the URL path, where project name can be empty. But when creating a project, name cannot be empty. But in both cases, I am preparing a model to create or delete the entity and hence this way

sunnyagg: should be have different validators for different operations than combining into one method

internal/subscription/core.go
```diff
@@ -72,3 +75,20 @@ func (c *Core) Exists(ctx context.Context, key string) (bool, error) {
 	}
 	return ok, nil
 }
+
+// DeleteSubscription deletes a subscription
+func (c *Core) DeleteSubscription(ctx context.Context, m *Model) error {
```
sunnyagg, line 80: should we also validate if the subscription belongs to a project as well

Abhishekvrshny: This is being checked here. https://github.com/razorpay/metro/pull/44/files#diff-62bfaba279adf117168ffb1d9690cc9c5d7227ad67cd181eb9934b4330bdf779R87 
key() returns the full consul path registry/projects/<project>/subscriptions/<subscription>

internal/project/validation.go
Abhishekvrshny: done

Abhishekvrshny:
@sunnyagg addressed the review comments

sunnyagg (approved):

PR #45, opened 2021-01-11
adds broker store to init and maintain clients

shivasishdas:
- Delegates broker-client init to `BrokerStore`. Any service needing a new `producer`, `consumer`, `admin` client will request it from the brokerStore.
- `BrokerStore` will manage the lifecycle of each client.

TODO : 
- Refactor configs. Currently, the registry config is duplicated inside each component
- Add tests


internal/brokerstore/brokerstore.go
```diff
@@ -0,0 +1,178 @@
...
+	return fmt.Sprintf("%v-%v", key.instanceType, key.topic)
+}
+
+// BrokerStore holds broker clients
+type BrokerStore struct {
+	// stores active producer clients for a key.
+	producerMap map[*Key]messagebroker.Producer
+
+	// stores active producer clients for a key
+	consumerMap map[*Key]messagebroker.Consumer
```
Abhishekvrshny, line 49: should these be sync maps?

Abhishekvrshny:
There is a change in proto too which is committed. Is that desired?

service/producer/publisherserver.go
```diff
@@ -12,23 +14,28 @@ import (
...
 }
 
 // Produce messages to a topic
 func (s publisherServer) Publish(ctx context.Context, req *metrov1.PublishRequest) (*metrov1.PublishResponse, error) {
 
 	log.Println("produce request received")
 
+	producer, err := s.brokerStore.GetExistingOrCreateProducer(ctx, messagebroker.ProducerClientOptions{Topic: req.Topic})
+	if err != nil {
+		return nil, err
```
Abhishekvrshny, line 32: merror.ToGRPCError(err)

internal/brokerstore/brokerstore.go
```diff
@@ -0,0 +1,178 @@
...
+	CreateConsumer(context.Context, messagebroker.ConsumerClientOptions) (messagebroker.Consumer, error)
+
+	// GetExistingOrCreateConsumer returns for an existing consumer instance, if available returns that else creates as new instance
+	GetExistingOrCreateConsumer(context.Context, messagebroker.ConsumerClientOptions) (messagebroker.Consumer, error)
+
+	// CreateProducer returns a new producer instance of the desired broker
+	CreateProducer(context.Context, messagebroker.ProducerClientOptions) (messagebroker.Producer, error)
+
+	// GetExistingOrCreateProducer returns for an existing producer instance, if available returns that else creates as new instance
+	GetExistingOrCreateProducer(context.Context, messagebroker.ProducerClientOptions) (messagebroker.Producer, error)
```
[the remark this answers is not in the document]
shivasishdas: flag to merge CreateProducer and GetExistingOrCreateProducer?

internal/brokerstore/brokerstore.go
shivasishdas: Was thinking of syncing using `consul lock` as `sync.Map` does not maintain types and accepts only `interface{}`. Anyways changed for now.

sunnyagg: we have seperate producer, consumer and admin maps, do we need to store type information in key as well ?

shivasishdas: That's to make the log readable. Else it will only be a topic name.

internal/brokerstore/brokerstore.go
```diff
@@ -0,0 +1,178 @@
...
+type InstanceType string
+
+const (
+	producer InstanceType = "producer"
+	consumer InstanceType = "consumer"
+	admin    InstanceType = "admin"
+)
+
+// Key ...
+type Key struct {
```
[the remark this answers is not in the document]
shivasishdas: - multiple consumers will operate on the same topic. we would differentiate using this key.
- better logging.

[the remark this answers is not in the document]
shivasishdas: define the key as => topic-partition-number

Abhishekvrshny (approved):

PR #49, opened 2021-01-13
node registration with registry

sunnyagg:
- Adds Node registration with registry
- Re write of leader election

### DIFF
--- a/pkg/leaderelection/candidate.go
+++ b/pkg/leaderelection/candidate.go
@@ -2,29 +2,28 @@ package leaderelection
 
 import (
 	"context"
+	"fmt"
 	"time"
+
+	"golang.org/x/sync/errgroup"
 
 	"github.com/razorpay/metro/pkg/logger"
 	"github.com/razorpay/metro/pkg/registry"
-	"k8s.io/apimachinery/pkg/util/wait"
-)
-
-const (
-	// JitterFactor used in wait utils
-	JitterFactor = 1.2
 )
 
 // New creates a instance of LeaderElector from a LeaderElection Config
-func New(c Config, registry registry.IRegistry) (*Candidate, error) {
+func New(id string, c Config, registry registry.IRegistry) (*Candidate, error) {
 	if err := c.Validate(); err != nil {
 		return nil, err
 	}
 
 	le := Candidate{
+		ID:       id,
 		config:   c,
 		registry: registry,
 		nodeID:   "",
 		leader:   false,
+		errCh:    make(chan error),
 	}
 
 	return &le, nil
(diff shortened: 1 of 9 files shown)


shivasishdas:
@sunnyagg Quick question. The implementation we wrote how different is it from the one mentioned here: https://clivern.com/leader-election-with-consul-and-golang/ ?

pkg/leaderelection/candidate.go
```diff
@@ -43,94 +50,106 @@ type Candidate struct {
...
-	}, c.config.RenewDeadline, leadCtx.Done())
+	var err error
 
-	// context returned done, release the lease
-	logger.Ctx(leadCtx).Info("context done, releasing lease")
-	c.release(leadCtx)
-	return leadCtx.Err()
-}
+	logger.Ctx(ctx).Info("registering node with registry")
+	c.nodeID, err = c.registry.Register(c.config.Name, c.config.LeaseDuration)
```
Abhishekvrshny, line 64: This register is required because once the node goes away, the session would become invalid and all nodes it created in consul would be deleted? Is that understanding correct?

Abhishekvrshny: should it be on config name or id then?

pkg/leaderelection/candidate.go
```diff
@@ -43,94 +50,106 @@ type Candidate struct {
...
+		logger.Ctx(ctx).Errorw("node registering failed with error", "error", err.Error())
+		return err
+	}
 
-// IsLeader returns true if the last observed leader was this client else returns false.
-func (c *Candidate) IsLeader() bool {
-	return c.leader
-}
+	logger.Ctx(ctx).Infow("acquiring node key", "key", c.config.NodePath)
+	acquired := c.registry.Acquire(c.nodeID, c.config.NodePath, time.Now().String())
```
Abhishekvrshny, line 71: why is this acquire required here?

Abhishekvrshny (approved):

pkg/leaderelection/candidate.go
sunnyagg: Yes, Assuming by node you mean keys created and acquired under registry.
in our case we are acquiring 2 
/registry/nodes/<id>
/leader/election

pkg/leaderelection/candidate.go
sunnyagg: if we dont acquire and just create the key, key wont be deleted on session expiry. we need to acquire to make it ephemeral

pkg/leaderelection/candidate.go
sunnyagg: >should it be on config name or id then?
On register, consul issues a session id, we can pass both name and id, for now passing id is optional as we dont have any use case, can be added later if required

PR #52, opened 2021-01-18
adding pulsarctl for admin

shivasishdas:


pkg/messagebroker/models.go
```diff
@@ -30,6 +34,7 @@ type CommitOnTopicRequest struct {
 	Topic     string
 	Partition int32
 	Offset    int64
+	ID        string
```
Abhishekvrshny, line 37: This message ID right? So, commit on kafka is per offset, but for pulsar per message? Can it be offset based in Pulsar too? If not, have separate methods?

pkg/messagebroker/models.go
```diff
@@ -8,7 +10,9 @@ type CreateTopicRequest struct {
 
 // DeleteTopicRequest ...
 type DeleteTopicRequest struct {
-	Name string
+	Name         string
+	Force        bool
+	NonPartioned bool
```
Abhishekvrshny, line 15: Add a comment saying `Force` and `NonPartioned ` are only requited for pulsar and ignored for kafka

Abhishekvrshny (approved):

PR #54, opened 2021-02-06
adds basic pubsub interface and its implementation

Abhishekvrshny:
- [X] implements high level pub/sub interface
- [X] implements streaming pull rpc
- [X] fixes load and store in brokerstore

TODO:
- [x] there are issues with concurrent pull calls. Needs to be solved.
- [ ] the pub/subs structure is still evolving. Refactoring is required


service/web/subscriberserver.go
```diff
@@ -34,20 +39,115 @@ func (s subscriberserver) CreateSubscription(ctx context.Context, req *metrov1.S
...
+			if err != nil {
+				logger.Ctx(ctx).Infow("error received from client", "msg", err.Error())
+				return nil
+			}
+		case req = <-reqChan:
+			// reset stream ack deadline seconds
+			timeout.Stop()
+			timeout = time.NewTicker(time.Duration(req.StreamAckDeadlineSeconds) * time.Second)
+			// ping request received with empty payload
+			if req.Subscription == "" {
```
shivasishdas, line 108: - should we add some validations on the pull request since `Subscription` is sent only when establishing a stream, thereafter only `AckIds` or `ModAckIds` will be sent. And an empty `Subscription` is allowed only when non-empty `AckIds` and `ModAckIds` are in the payload.

- Maybe another validation when an `ackId` or `modAck` Id comes in as the first request.

service/web/subscriberserver.go
```diff
@@ -34,20 +39,115 @@ func (s subscriberserver) CreateSubscription(ctx context.Context, req *metrov1.S
...
+				continue
+			}
+			ackReq, ModAckReq := subscriber.FromProto(req)
+			// if its the first req and subscriber is not yet initialised
+			if pullStream == nil {
+				pullStream, err = newPullStream(ctx,
```
shivasishdas, line 114: how is the number of messages to be read in the stream controlled? Don't see any such param under [StreamingPullRequest](https://github.com/razorpay/metro-proto/blob/ae77c2506dc966865d77aca620b98768f7ddeba3/metro/proto/v1/spec.proto#L620)

service/web/stream.go
```diff
@@ -0,0 +1,58 @@
...
+	return s.subscriptionSubscriber.ModifyAckDeadline(ctx, req)
+}
+
+func (s pullStream) stop() {
+	s.cancelFunc()
+	<-s.stopChan
+}
+
+func newPullStream(ctx context.Context, clientID string, subscription string, subscriberCore subscriber.ICore, responseChan chan<- metrov1.PullResponse) (*pullStream, error) {
+	subs, err := subscriberCore.NewSubscriber(ctx, clientID, subscription, 10, 0, 0)
```
shivasishdas, line 52: passing the server `ctx` directly to the `subscriber`. Should we make a copy `context.WithCancel(ctx)`?

service/web/subscriberserver.go
```diff
@@ -34,20 +39,115 @@ func (s subscriberserver) CreateSubscription(ctx context.Context, req *metrov1.S
...
+				return nil
+			}
+			if err != nil {
+				logger.Ctx(ctx).Infow("error received from client", "msg", err.Error())
+				return nil
+			}
+		case req = <-reqChan:
+			// reset stream ack deadline seconds
+			timeout.Stop()
+			timeout = time.NewTicker(time.Duration(req.StreamAckDeadlineSeconds) * time.Second)
```
shivasishdas, line 106: should this be moved to the end (or wrapped in `defer`) after the current request processing completes?

shivasishdas (changes requested):

service/web/subscriberserver.go
```diff
@@ -34,20 +39,115 @@ func (s subscriberserver) CreateSubscription(ctx context.Context, req *metrov1.S
...
 }
 
 // StreamingPull ...
 func (s subscriberserver) StreamingPull(server metrov1.Subscriber_StreamingPullServer) error {
-	// TODO: Implement
+	var pullStream *pullStream
+	var req *metrov1.StreamingPullRequest
+	var timeout *time.Ticker
+	var err error
+	timeout = time.NewTicker(5 * time.Second) // init with some sane value
```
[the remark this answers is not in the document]
Abhishekvrshny: `req` is not yet initialized. It is being initialised with some sane value and reset later once `req` is known.

service/web/subscriberserver.go
Abhishekvrshny: Yep, I guess this server controlle. I am thinking of defining Max outstanding messages and using that here too.

service/web/subscriberserver.go
Abhishekvrshny: Ack

service/web/stream.go
Abhishekvrshny: Ack

shivasishdas (approved):
LGTM

internal/brokerstore/brokerstore.go
```diff
@@ -81,55 +79,54 @@ func NewBrokerStore(variant string, config *messagebroker.BrokerConfig) (IBroker
...
 }
 
 // GetOrCreateProducer returns for an existing producer instance, if available returns that else creates as new instance
 func (b *BrokerStore) GetOrCreateProducer(ctx context.Context, op messagebroker.ProducerClientOptions) (messagebroker.Producer, error) {
-	key := NewKey(b.variant, op.Partition)
-
-	if producer, exists := b.producerMap.Load(key.String()); exists {
-		return producer.(messagebroker.Producer), nil
-	}
+	key := NewKey(b.variant, "")
```
sunnyagg, line 116: we seem to be creating a new producer client everytime, didn't got why, if that is require, do we need to keep a map and store them ?

internal/brokerstore/brokerstore.go
```diff
@@ -81,55 +79,54 @@ func NewBrokerStore(variant string, config *messagebroker.BrokerConfig) (IBroker
...
+		partitionLock: partitionlocker.NewPartitionLocker(&sync.Mutex{}),
 	}, nil
 }
 
 // GetOrCreateConsumer returns for an existing consumer instance, if available returns that else creates as new instance
-func (b *BrokerStore) GetOrCreateConsumer(ctx context.Context, op messagebroker.ConsumerClientOptions) (messagebroker.Consumer, error) {
-	key := NewKey(b.variant, op.Partition)
-
-	if consumer, exists := b.consumerMap.Load(key.String()); exists {
+func (b *BrokerStore) GetOrCreateConsumer(ctx context.Context, id string, op messagebroker.ConsumerClientOptions) (messagebroker.Consumer, error) {
```
sunnyagg, line 89: should we just rename this to GetConsumer, GetProducer ?

internal/publisher/core.go
```diff
@@ -0,0 +1,59 @@
...
+		msg.MessageId = ""
+		msg.PublishTime = nil
+		// marshal proto with all attributes for publishing
+		dataWithMeta, err := proto.Marshal(msg)
+		// TODO: check the scenario where one out of many messages fail in google pubsub
+		if err != nil {
+			return nil, fmt.Errorf("unable to marshal message")
+		}
+		// TODO: rationalise TimeoutSec
+		msgResp, err := producer.SendMessages(ctx, messagebroker.SendMessageToTopicRequest{
```
sunnyagg, line 45: sendMessages takes single message, possible to call it SendMessage ?

internal/publisher/core.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		logger.Ctx(ctx).Errorw("error in getting producer", "msg", err.Error())
+		return nil, err
+	}
+
+	msgIDs := make([]string, 0)
+
+	for _, msg := range req.Messages {
+		// unset message id and publishtime if set
+		msg.MessageId = ""
```
sunnyagg, line 36: do we allow setting these in future ? if not, should we validate this and dont allow setting them at all

internal/subscriber/core.go
```diff
@@ -0,0 +1,54 @@
...
+	subscriptionCore subscription.ICore
+}
+
+// NewCore returns a new core
+func NewCore(bs brokerstore.IBrokerStore, subscriptionCore subscription.ICore) *Core {
+	return &Core{bs, subscriptionCore}
+}
+
+// NewSubscriber initiates a new subscriber for a given topic
+func (c *Core) NewSubscriber(ctx context.Context, id string, subscription string, timeoutInSec int, maxOutstandingMessages int64, maxOutstandingBytes int64) (ISubscriber, error) {
```
sunnyagg, line 30: can we pass subscription object itself here than all params, most params seems like sub config

service/web/subscriberserver.go
```diff
@@ -34,20 +39,120 @@ func (s subscriberserver) CreateSubscription(ctx context.Context, req *metrov1.S
...
+				return merror.ToGRPCError(err)
+			}
+			// TODO: check and add if error is returned when subscription is changed in subsequent requests
+			// continue on ping request
+			if ackReq.IsEmpty() && modAckReq.IsEmpty() {
+				continue
+			}
+			// if its the first req and subscriber is not yet initialised
+			if pullStream == nil {
+				if req.Subscription == "" {
```
sunnyagg, line 117: what if the requested subscription is push sub ? 
what should happen to a active pullStream if a sub i converted to push sub later ?

sunnyagg (approved):

internal/publisher/core.go
Abhishekvrshny: Goofle pubsub doesn't throw an error here and silently ignores the user set message id.

internal/brokerstore/brokerstore.go
Abhishekvrshny: Fixed. Added a comment to review this while doing perf

internal/brokerstore/brokerstore.go
Abhishekvrshny: Done

internal/publisher/core.go
Abhishekvrshny: Done

internal/subscriber/core.go
Abhishekvrshny: should not pass the proto subscription, because this layer should not understand proto. Also, not in favour of creating encapsulating structs for args as they pose additional overhead of validating all required options while passing them inline makes it explicit

service/web/subscriberserver.go
Abhishekvrshny: good point. Added this TODO:
`TODO: check if the requested subscription is push based and handle it the way pubsub does`

PR #59, opened 2021-03-01
ack nack

shivasishdas:


service/web/subscriberserver.go
```diff
@@ -62,17 +63,60 @@ func (s subscriberserver) Pull(ctx context.Context, req *metrov1.PullRequest) (*
...
+			timeout = time.NewTicker(time.Duration(req.StreamAckDeadlineSeconds) * time.Second)
+			parsedReq, parseErr := newParsedStreamingPullRequest(req)
+			if parseErr != nil {
+				logger.Ctx(ctx).Errorw("error is parsing pull request", "request", req, "error", parseErr.Error())
+				return nil
+			}
+
+			// request to init a new stream
+			if parsedReq.HasSubscription() {
+				err := s.psm.CreateNewStream(server, parsedReq)
```
Abhishekvrshny, line 89: this will get created on every syream request?

Abhishekvrshny:
Fixed streaming pull with this commit id ea06d0c3a778bf350bd82a89dc9f5711bf1b582f. 

Note:
- ModAck definition in subscriber.go needs to be fixed
- Its implementation also needs to be fixed in all layers
- ModAck with deadline = 0 means Nack, and we might only need to support only this for now (ref: https://github.com/googleapis/google-cloud-go/blob/pubsub/v1.10.0/pubsub/iterator.go#L348)

PR #60, opened 2021-03-03
Node registration, nodes and subs List APIS

sunnyagg:
- Adds Nodes model
- Adds support for node registration
- Adds Repo fetch for nodes/subscription Keys


Abhishekvrshny (approved):

service/worker/service.go
```diff
@@ -69,17 +81,47 @@ func (c *Service) Start() error {
...
+		return err
+	}
+
+	// create nodeCache
+	c.nodeCache, err = nodeCore.ListKeys(c.ctx)
+	if err != nil {
+		return err
+	}
+
+	c.subCache, err = subscriptionCore.ListKeys(c.ctx)
```
Abhishekvrshny, line 115: should this happen on all nodes or only on the one which is the leader?

sunnyagg: right, will move to leader func before setting up watch

PR #62, opened 2021-03-08
metro scheduler implementation

sunnyagg:
- Adds nodebinding model
- watch handlers for node, subscription, binding changes
- Scheduler implementation (Load Based, Random)
- Supporting Registry, Repo, Core APIs

### DIFF
--- a/service/worker/service.go
+++ b/service/worker/service.go
@@ -52,7 +52,7 @@ type Service struct {
 	nodeCache        []*node.Model
 	subCache         []*subscription.Model
 	nodebindingCache []*nodebinding.Model
-	pushHandlers     map[string]pushsubscriber
+	pushHandlers     map[string]*pushsubscriber
 	scheduler        *scheduler.Scheduler
 }
 
@@ -70,6 +70,7 @@ func NewService(ctx context.Context, workerConfig *Config, registryConfig *regis
 		nodeCache:        []*node.Model{},
 		subCache:         []*subscription.Model{},
 		nodebindingCache: []*nodebinding.Model{},
+		pushHandlers:     map[string]*pushsubscriber{},
 	}
 }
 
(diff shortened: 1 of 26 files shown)


service/worker/service.go
```diff
@@ -245,14 +266,14 @@ func (svc *Service) lead(ctx context.Context) error {
...
 		return err
 	}
 
 	// create subscription cache
 	logger.Ctx(ctx).Infof("getting active subscriptions for cache")
-	svc.subCache, err = svc.subscriptionCore.ListKeys(svc.ctx)
+	svc.subCache, err = svc.subscriptionCore.List(svc.ctx, "")
 	if err != nil {
 		return err
 	}
```
Abhishekvrshny, line 279: I think there can be a race condition here, i.e a node or subs can be added or deleted between this step and when the actual watch is set. Can we update these caches as part of setting the watch itself?

service/worker/service.go
```diff
@@ -327,68 +346,169 @@ func (svc *Service) stepDown() {
...
+			if old.Key() == new.Key() {
 				found = true
 				break
 			}
 		}
 
 		if !found {
-			logger.Ctx(svc.ctx).Infow("node added", "node_key", new)
+			logger.Ctx(svc.ctx).Infow("node added", "key", new.Key())
+			// Do nothing for now
```
Abhishekvrshny, line 511: This should also trigger scheduler right?

service/worker/service.go
```diff
@@ -280,8 +300,7 @@ func (svc *Service) lead(ctx context.Context) error {
 		WatchPath: "metro/subscriptions",
 		Handler: func(ctx context.Context, pairs []registry.Pair) {
 			logger.Ctx(ctx).Infow("nodes watch handler data", "pairs", pairs)
-			newSubs := registry.GetKeys(pairs)
-			svc.handleSubUpdates(newSubs)
+			svc.handleSubUpdates(pairs)
```
Abhishekvrshny, line 303: `handleNodeUpdates ` and `handleSubUpdates ` happens concurrently. Do you see any side-effects? Does this need to be synchronized?

service/worker/service.go
sunnyagg: when we create watch, that gives a mandatory first callback with all data present under that key prefix, i guess we can then just remove the manual fetch and depend only on the watch, as first watch will update the empty cache

service/worker/service.go
sunnyagg: for now, rebalance is not implemented, we dont need to change the current assignment for node additions

service/worker/service.go
sunnyagg: concurrently is fine, issue is with the failures in both cases, as watch is triggered only once, we need some intermediate storage if processing fails

i.e. if a subscription scheduling fails, put that in some channel and process it in separate go routine

Abhishekvrshny (approved):

PR #64, opened 2021-03-10
ack nack impl

shivasishdas:


internal/subscriber/subscriber.go
```diff
@@ -12,96 +16,369 @@ import (
...
+	// do a deadline based eviction for all active topic-partition heaps
+	for _, metadata := range s.consumedMessageStats {
+
+		deadlineBasedHeap := metadata.deadlineBasedMinHeap
+		if deadlineBasedHeap.IsEmpty() {
+			return
+		}
+
+		// peek deadline heap
+		peek := deadlineBasedHeap.Indices[0]
```
Abhishekvrshny, line 214: noticed panic here. You might want to check length of the array here.

shivasishdas: Already handled at : https://github.com/razorpay/metro/pull/64/files#diff-4bcad8e79804b094286e6698fe58e6f08d402b61a217a004fd4b078034ee9992R209
We deployed an older commit.

Abhishekvrshny (approved):

PR #66, opened 2021-03-11
adds push stream webhook handling

sunnyagg:
- Push stream implementation (Reads data from broker, publish to a channel, read data from response channel in another go routine, fire webhooks and ack/nack)


service/worker/stream.go
```diff
@@ -0,0 +1,156 @@
...
+				res := <-ps.subs.GetResponseChannel()
+				logger.Ctx(ps.ctx).Infow("writing subscriber data to channel", "res", res)
+				ps.responseChan <- res
+			}
+		}
+		logger.Ctx(ps.ctx).Infow("returning from pull stream go routine")
+		return nil
+	})
+
+	errGrp.Go(func() error {
```
Abhishekvrshny, line 85: Do we need 2 goroutines here?

sunnyagg: for end to end processing 1 can do, but for error handling this looked better, 1 go routine to read and and any error in this will require similar kind of handling and 1 go routine to push to subscriber and should have similar kind of error handling on subscriber being down etc

Abhishekvrshny (approved):

PR #84, opened 2021-03-22
metrics setup

shivasishdas:
- Setting up a few basic app metrics
- Setting web metric port to `9000` and worker metric port to `9001`. Refer [PR](https://github.com/razorpay/kube-manifests/pull/1018/files)


internal/subscriber/subscriber.go
```diff
@@ -157,6 +165,9 @@ func (s *Subscriber) acknowledge(ctx context.Context, req *AckMessage) {
 	}
 
 	removeMessageFromMemory(stats, req.MessageID)
+
+	subscriberMessagesAckd.WithLabelValues(s.topic, s.subscription).Add(1)
+	subscriberTimeTakenToAckMsg.WithLabelValues(s.topic, s.subscription).Observe(time.Now().Sub(msg.PublishTime).Seconds())
```
Abhishekvrshny, line 170: This is diff of the time when the message was published and then acked, right? Should we have a similar metric from publish to consume and from consume to ack/modack?

Abhishekvrshny: Also, similar metric for modack is missing

internal/subscriber/subscriber.go
```diff
@@ -340,13 +359,15 @@ func (s *Subscriber) Run(ctx context.Context) {
...
 						s.consumedMessageStats[tp].maxCommittedOffset = resp.Offset
+						subscriberMessagesConsumed.WithLabelValues(msg.Topic, s.subscription).Add(1)
 					}
 
 					ackDeadline := time.Now().Add(minAckDeadline).Unix()
 					s.consumedMessageStats[tp].Store(msg, ackDeadline)
 
 					ackID := NewAckMessage(s.subscriberID, msg.Topic, msg.Partition, msg.Offset, int32(ackDeadline), msg.MessageID).BuildAckID()
 					sm = append(sm, &metrov1.ReceivedMessage{AckId: ackID, Message: protoMsg, DeliveryAttempt: 1})
+					subscriberMemoryMessagesCountTotal.WithLabelValues(s.topic, s.subscription).Set(float64(len(s.consumedMessageStats[tp].consumedMessages)))
```
Abhishekvrshny, line 370: this is the current total number of messages in the heap right?

Abhishekvrshny:
should an `env` label be also added? eg `stage`, `prod` etc?

Abhishekvrshny:
Add GRPC request metrics with latency and kafka, consul request metrics with latency

internal/node/core.go
```diff
@@ -30,6 +31,11 @@ func NewCore(repo IRepo) *Core {
 
 // CreateNode creates a new node
 func (c *Core) CreateNode(ctx context.Context, m *Model) error {
+	nodeOperationCount.WithLabelValues(env, "CreateNode").Inc()
+
+	startTime := time.Now()
+	defer nodeOperationTimeTaken.WithLabelValues(env, "CreateNode").Observe(time.Now().Sub(startTime).Seconds())
```
sunnyagg, line 37: time in ms than secs

shivasishdas: Going by this : https://prometheus.io/docs/practices/naming/
> ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers). See below for a list of base units.

internal/publisher/core.go
```diff
@@ -55,6 +55,7 @@ func (p *Core) Publish(ctx context.Context, req *metrov1.PublishRequest) ([]stri
 		}
 		msgIDs = append(msgIDs, msgResp.MessageID)
 	}
+	publisherMessagesPublished.WithLabelValues(env, req.Topic).Add(float64(len(msgIDs)))
```
sunnyagg, line 58: why is this float ?

sunnyagg (approved):

internal/publisher/core.go
shivasishdas: The function expects `float64` 

```
// To create Counter instances, use NewCounter.
type Counter interface {
	Metric
	Collector

	// Inc increments the counter by 1. Use Add to increment it by arbitrary
	// non-negative values.
	Inc()
	// Add adds the given value to the counter. It panics if the value is <
	// 0.
	Add(float64)
}
```

internal/node/core.go
shivasishdas: changed

sunnyagg (approved):

PR #96, opened 2021-04-14
adding dlq push

shivasishdas:
- [X] On exhausting retries, push message to dlq

TODO
- [ ] figure out message replay semantics from dlq

### DIFF
--- a/internal/topic/model.go
+++ b/internal/topic/model.go
@@ -11,6 +11,9 @@ const (
 	// RetryTopicSuffix every primary topic will have a retry topic with this suffix as well
 	RetryTopicSuffix = "-retry"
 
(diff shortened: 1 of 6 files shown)


internal/topic/model.go
```diff
@@ -23,6 +26,7 @@ type Model struct {
 	ExtractedProjectID string
 	ExtractedTopicName string
 	RetryTopicName     string
+	DLQTopicName       string
```
sunnyagg, line 29: update test files as well ?

sunnyagg (approved):

internal/topic/model.go
shivasishdas: added

PR #101, opened 2021-04-14
proxy request incase of originating server address mismatch

shivasishdas:
**Message Ack stickiness due to Unary Ack / Nack / ModAck**

Ack/Nack/ModAck requests for messages can be load-balanced and land on a server that is not holding the message in memory. We need to proxy the message to the server from where it was delivered.

### DIFF
--- a/service/web/stream/manager.go
+++ b/service/web/stream/manager.go
@@ -180,6 +180,9 @@ func (s *Manager) ModifyAcknowledgement(ctx context.Context, parsedReq *ParsedSt
 // proxy request to the specified server
 func (pr *proxyRequest) do(ctx context.Context) {
 
+	// from config grpcServerAddr would be read as => 0.0.0.0:8081
+	// we extract the port(8081) out of the host+grpc_port string and append it to the ack_msg_addr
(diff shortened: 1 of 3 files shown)


Abhishekvrshny:
Can the proxy layer be common in those 2 functions?

Abhishekvrshny (approved):

service/web/stream/manager.go
```diff
@@ -107,23 +109,69 @@ func (s *Manager) CreateNewStream(server metrov1.Subscriber_StreamingPullServer,
...
+			proxyAddr := ackMsg.ServerAddress
+			if _, ok := msgsToBeProxied[proxyAddr]; !ok {
+				// init empty slice
+				msgsToBeProxied[proxyAddr] = make([]*subscriber.AckMessage, 0)
+			}
+			msgsToBeProxied[proxyAddr] = append(msgsToBeProxied[proxyAddr], ackMsg)
 		}
 	}
 
+	if len(msgsToBeProxied) > 0 {
```
sunnyagg, line 133: can we refactor this part into a common function, seems duplicated below

shivasishdas: done.

PR #105, opened 2021-04-23
subscription based retry and dlq topics

sunnyagg:
- Moves retry and dlq support at subscription level
- Topic names cannot have -retry and -dlq suffix as they are being used internally
- Retry topics can't have direct subscriptions

### DIFF
--- a/internal/subscription/core.go
+++ b/internal/subscription/core.go
@@ -71,9 +71,7 @@ func (c *Core) CreateSubscription(ctx context.Context, m *Model) error {
 
 	var topicModel *topic.Model
 	if topicModel, err = c.topicCore.Get(ctx, m.GetTopic()); err != nil {
-		if err != nil {
-			return err
-		}
+		return err
 	}
 
 	// for subscription over deadletter topics, skip the retry and deadletter topic creation
@@ -93,7 +91,7 @@ func (c *Core) CreateSubscription(ctx context.Context, m *Model) error {
 		}
 
 		// create deadletter topic for subscription
-		// TODO: read the deadletter policy and update occordingly
+		// TODO: read the deadletter policy and update accordingly
 		err = c.topicCore.CreateDeadLetterTopic(ctx, &topic.Model{
 			Name:               m.GetDeadLetterTopic(),
 			ExtractedTopicName: m.ExtractedSubscriptionName + topic.DeadLetterTopicSuffix,
@@ -102,7 +100,7 @@ func (c *Core) CreateSubscription(ctx context.Context, m *Model) error {
 		})
 
 		if err != nil {
-			logger.Ctx(ctx).Errorw("failed to create deadletter topic for subscription", "name", m.GetDeadLetterTopic(), "error" , err.Error())
+			logger.Ctx(ctx).Errorw("failed to create deadletter topic for subscription", "name", m.GetDeadLetterTopic(), "error", err.Error())
 			return err
 		}
 	}
@@ -198,7 +196,7 @@ func (c *Core) List(ctx context.Context, prefix string) ([]*Model, error) {
 
 	prefix = common.GetBasePrefix() + prefix
 
-	out := []*Model{}
+	var out []*Model
 	ret, err := c.repo.List(ctx, prefix)
 	if err != nil {
 		return nil, err
--- a/internal/subscription/core.go
+++ b/internal/subscription/core.go
@@ -75,11 +75,10 @@ func (c *Core) CreateSubscription(ctx context.Context, m *Model) error {
 	}
 
 	var topicModel *topic.Model
-	if topicModel, err = c.topicCore.Get(ctx, m.Topic); !ok {
+	if topicModel, err = c.topicCore.Get(ctx, m.Topic); err != nil {
 		if err != nil {
 			return err
 		}
-		return merror.Newf(merror.NotFound, "topic with name %s not found", m.Topic)
 	}
 
 	admin, aerr := c.brokerStore.GetAdmin(ctx, messagebroker.AdminClientOptions{})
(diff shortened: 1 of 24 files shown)


internal/subscription/validation.go
```diff
@@ -55,6 +55,8 @@ func getValidatedModel(ctx context.Context, req *metrov1.Subscription) (*Model,
 	m.Labels = req.GetLabels()
 	m.ExtractedSubscriptionProjectID = p
 	m.ExtractedSubscriptionName = s
+	m.RetryTopic = topic.GetTopicName(p, s+topic.RetryTopicSuffix)
```
shivasishdas, line 58: Would be good if we add a comment here explaining why we are using the subscription name and not the topic name to create retry and dlq topics.

internal/subscription/model.go
```diff
@@ -14,6 +14,8 @@ type Model struct {
 	common.BaseModel
 	Name                           string
 	Topic                          string
+	RetryTopic                     string
```
shivasishdas, line 17: Since these are now part of the subscription model it will get persisted this to consul as well. Not sure if this is needed. Alternative would be to just have `subscription.GetRetryTopicName(...)` and `subscription.GetDLQTopicName(...)` and construct the names on the fly whenever needed.

internal/subscription/validation.go
```diff
@@ -55,6 +55,8 @@ func getValidatedModel(ctx context.Context, req *metrov1.Subscription) (*Model,
...
+	m.DeadLetterTopic = topic.GetTopicName(p, s+topic.DLQTopicSuffix)
```
shivasishdas, line 59:
```suggestion
	m.DeadLetterTopic = topic.GetTopicName(p, s+topic.DeadLetterTopicSuffix)
```
nit: let's keep the naming consistent by using either `DeadLetter` or `DLQ` everywhere.

internal/subscription/core.go
```diff
@@ -204,6 +219,8 @@ func (c *Core) Get(ctx context.Context, key string) (*Model, error) {
 	}
 	prefix := common.GetBasePrefix() + Prefix + projectID + "/" + subscriptionName
 
+	logger.Ctx(ctx).Infow("fetching subscription", "key", prefix)
```
shivasishdas, line 222: was this a test logger to be removed later?

internal/topic/core.go
```diff
@@ -58,8 +58,7 @@ func (c *Core) CreateTopic(ctx context.Context, m *Model) error {
 	if err != nil {
 		return err
 	}
-	// TODO: take number of partitions as input
-	_, err = admin.CreateTopic(ctx, messagebroker.CreateTopicRequest{m.Name, 2})
+	_, err = admin.CreateTopic(ctx, messagebroker.CreateTopicRequest{m.Name, m.NumPartitions})
```
shivasishdas, line 61: Need to add validations around `m.NumPartitions` value. 

Just as we have a `DefaultNumPartitions ` value, I think we should define a `MaxAllowedNumPartitions ` value as well to restrict accidental over-provisioning. Thoughts?

shivasishdas (changes requested):

internal/subscription/core.go
sunnyagg: was added as test log, but thought of keeping it for debugging

internal/subscription/model.go
sunnyagg: Yes, was thinking on this, makes sense for RetryTopic since that is internal. DLQ will still be exposed so should be fine to keep

internal/subscription/core.go
```diff
@@ -68,12 +71,38 @@ func (c *Core) CreateSubscription(ctx context.Context, m *Model) error {
...
+
+	admin, aerr := c.brokerStore.GetAdmin(ctx, messagebroker.AdminClientOptions{})
+	if aerr != nil {
+		return aerr
+	}
+
+	// create retry topic for subscription
+	_, terr := admin.CreateTopic(ctx, messagebroker.CreateTopicRequest{
+		Name:          m.RetryTopic,
+		NumPartitions: topicModel.NumPartitions,
```
shivasishdas, line 90: @Abhishekvrshny @sunnyagg Should `retry` and `dlq` topics be also created with the same number of partitions as the primary topic?

shivasishdas (changes requested):

internal/subscription/core.go
sunnyagg: since we listen to retry topic for every steam, we will end up spawning n consumers for retry topic too, i think its better to keep it same. dlq is still debatable, in future we can give control to user for dlq.

internal/subscription/validation.go
sunnyagg: adding to the model

internal/topic/core.go
sunnyagg: Yes, we need a lot of validations here, creating a [issue](https://github.com/razorpay/metro/issues/108) for this to fix in separate PR

internal/subscription/core.go
```diff
@@ -68,12 +68,45 @@ func (c *Core) CreateSubscription(ctx context.Context, m *Model) error {
...
+		// TODO: read the deadletter policy and update occordingly
+		err = c.topicCore.CreateDeadLetterTopic(ctx, &topic.Model{
+			Name:               m.GetDeadLetterTopic(),
+			ExtractedTopicName: m.ExtractedSubscriptionName + topic.DeadLetterTopicSuffix,
+			ExtractedProjectID: m.ExtractedTopicProjectID,
+			NumPartitions:      topicModel.NumPartitions,
+		})
+
+		if err != nil {
+			logger.Ctx(ctx).Errorw("failed to create deadletter topic for subscription", "name", m.GetDeadLetterTopic())
```
shivasishdas, line 105:
```suggestion
			logger.Ctx(ctx).Errorw("failed to create deadletter topic for subscription", "name", m.GetDeadLetterTopic(), "error" , err.Error())
```
log the actual error as well. Same above.

shivasishdas (approved):

internal/subscription/core.go
```diff
@@ -68,12 +68,45 @@ func (c *Core) CreateSubscription(ctx context.Context, m *Model) error {
...
+		// TODO: update based on retry policy
+		err = c.topicCore.CreateRetryTopic(ctx, &topic.Model{
+			Name:               m.GetRetryTopic(),
+			ExtractedTopicName: m.ExtractedSubscriptionName + topic.RetryTopicSuffix,
+			ExtractedProjectID: m.ExtractedTopicProjectID,
+			NumPartitions:      topicModel.NumPartitions,
+		})
+
+		if err != nil {
+			logger.Ctx(ctx).Errorw("failed to create retry topic for subscription", "name", m.GetRetryTopic())
```
sunnyagg, line 91:
```suggestion
			logger.Ctx(ctx).Errorw("failed to create retry topic for subscription", "name", m.GetRetryTopic(), "error", err.Error())
```

PR #117, opened 2021-05-10
refactor channel hops

shivasishdas:
- removes `responseChan` hop from `PushStream` in worker and directly listens on subscribers response chan.
- removes `deadlineTickerChan` from `Subscriber` and directly listens on the ticker channel.

### DIFF
--- a/service/worker/stream.go
+++ b/service/worker/stream.go
@@ -24,7 +24,6 @@ type PushStream struct {
 	subscriberCore   subscriber.ICore
 	subs             subscriber.ISubscriber
 	httpClient       http.Client
-	responseChan     chan metrov1.PullResponse
 	stopCh           chan struct{}
 	doneCh           chan struct{}
 }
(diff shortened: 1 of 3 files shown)


service/worker/stream.go
```diff
@@ -80,32 +76,18 @@ func (ps *PushStream) Start() error {
...
-	errGrp.Go(func() error {
-		// read from response channel and fire a webhook
-		for {
-			select {
-			case <-gctx.Done():
-				return gctx.Err()
-			default:
-				data := <-ps.responseChan
+				// wait for message to be available on the response channel
+				data := <-ps.subs.GetResponseChannel()
```
sunnyagg, line 82: read it in a separate case statement  ?

shivasishdas: Currently `ps.subs.GetResponseChannel()` is a blocking call. Every request will wait for a response. If I move it to a case, then items in requests chan can exceed response chan. Are we ok with this?

sunnyagg: we can control no of of requests by limiting the request and response channel sizes. 
keeping request response in processing in one block will make it slow, we should then consider hitting webhooks in separate go routines

PR #124, opened 2021-05-24
worker handlers shutdown fixes

shivasishdas:


service/worker/service.go
```diff
@@ -487,7 +504,12 @@ func (svc *Service) handleNodeBindingUpdates(ctx context.Context, newBindingPair
 				}
 			}(ctx)
 
-			svc.pushHandlers[newBinding.Key()] = handler
+			// store only if subscriber creation succeeds
+			error := <-notifyCh
```
sunnyagg, line 508: if start errors out, can we design stop in a way that we handle cases, we shouldn't need notify channel seperately

service/worker/service.go
```diff
@@ -447,19 +450,32 @@ func (svc *Service) handleNodeBindingUpdates(ctx context.Context, newBindingPair
...
 			}
 		}
 
 		if !found {
-			logger.Ctx(ctx).Infow("binding removed", "key", old.Key())
-			handler := svc.pushHandlers[old.Key()]
-			handler.Stop()
-			delete(svc.pushHandlers, old.Key())
+			func() {
+				svc.pushHandlersLock.Lock()
```
sunnyagg, line 464: instead of mutex can the datastructure itself be like that which supports synchronised read/writes

shivasishdas: yup. Involves a bit of refactoring. Will take ut up in subsequent PRs. Tracking [here](https://github.com/razorpay/metro/issues/126)

sunnyagg (approved):

PR #128, opened 2021-05-24
modify topic admin API

shivasishdas:
- Adding Admin ModifyTopic request
- Proto changes [here](https://github.com/razorpay/metro-proto/pull/18/files)

### DIFF
--- a/internal/topic/core.go
+++ b/internal/topic/core.go
@@ -71,7 +71,7 @@ func (c *Core) CreateTopic(ctx context.Context, m *Model) error {
 	}
 
 	// create regsitry entry
-	return c.repo.Create(ctx, m)
+	return c.repo.Save(ctx, m)
 }
 
 // CreateRetryTopic creates a retry topic for the given primary topic and name
@@ -89,7 +89,7 @@ func (c *Core) CreateDeadLetterTopic(ctx context.Context, model *Model) error {
 	}
 
 	// create registry entry
-	return c.repo.Create(ctx, model)
+	return c.repo.Save(ctx, model)
 }
 
 // UpdateTopic implements topic updation
@@ -108,7 +108,7 @@ func (c *Core) UpdateTopic(ctx context.Context, m *Model) error {
 		return merror.New(merror.NotFound, "topic not found")
 	}
 
-	return c.repo.Create(ctx, m)
+	return c.repo.Save(ctx, m)
 }
 
 // Exists checks if the topic exists with a given key
--- a/internal/topic/core.go
+++ b/internal/topic/core.go
@@ -15,6 +15,7 @@ import (
 // ICore is an interface over topic core
 type ICore interface {
 	CreateTopic(ctx context.Context, model *Model) error
+	UpdateTopic(ctx context.Context, model *Model) error
 	Exists(ctx context.Context, key string) (bool, error)
 	ExistsWithName(ctx context.Context, name string) (bool, error)
 	DeleteTopic(ctx context.Context, m *Model) error
@@ -79,7 +80,7 @@ func (c *Core) CreateRetryTopic(ctx context.Context, model *Model) error {
 	return c.createBrokerTopic(ctx, model)
 }
 
-// CreateDeadLetterTopic creates a deadletter topic for the given primary topic and name
+// CreateDeadLetterTopic creates a dead letter topic for the given primary topic and name
 func (c *Core) CreateDeadLetterTopic(ctx context.Context, model *Model) error {
 	// create broker topic
 	err := c.createBrokerTopic(ctx, model)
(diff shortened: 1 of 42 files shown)


service/web/adminserver.go
```diff
@@ -65,3 +69,46 @@ func (s adminServer) DeleteProject(ctx context.Context, req *metrov1.Project) (*
...
 	return &emptypb.Empty{}, nil
 }
+
+// ModifyTopic modify an existing topic
+func (s adminServer) ModifyTopic(ctx context.Context, req *metrov1.AdminTopic) (*emptypb.Empty, error) {
+
+	logger.Ctx(ctx).Infow("received admin request to modify topic",
+		"name", req.Name, "num_partitions", req.NumPartitions)
+
+	m, err := topic.GetValidatedAdminModel(ctx, req)
```
sunnyagg, line 79: should rename, GetValidatedTopicForAdminUpdate or something

pkg/messagebroker/kafka.go
```diff
@@ -571,3 +572,66 @@ func (k *KafkaBroker) Close(ctx context.Context) error {
 
 	return nil
 }
+
+// AddTopicPartitions adds partitions to an existing topic
```
sunnyagg, line 576: add comments here, this could be breaking if the topic is being used for ordered messages, should be used with caution :)

pkg/messagebroker/kafka.go
```diff
@@ -571,3 +572,66 @@ func (k *KafkaBroker) Close(ctx context.Context) error {
...
+	}
+
+	// metadata contains non-existent topic names as well with partition count as zero
+	currPartitionCount := len(metadata.Topics[tp].Partitions)
+	if currPartitionCount == 0 {
+		// invalid topic
+		return nil, fmt.Errorf("topic [%v] not found", tp)
+	}
+
+	if request.NumPartitions == currPartitionCount {
```
sunnyagg, line 606: this validation should be done via topic registry layer ?

pkg/messagebroker/messagebroker.go
```diff
@@ -16,6 +16,9 @@ type Admin interface {
 
 	// deletes an existing topic
 	DeleteTopic(context.Context, DeleteTopicRequest) (DeleteTopicResponse, error)
+
+	// adds partitions to an existing topic
+	AddTopicPartitions(context.Context, AddTopicPartitionRequest) (*AddTopicPartitionResponse, error)
```
sunnyagg, line 21: is this only add or we can use it to remove as well ?

shivasishdas: Kafka does not support decreasing partition counts on an existing topic.

pkg/messagebroker/kafka.go
shivasishdas: Since `messagebroker` is a `pkg` this function can ideally be called from anywhere without prior topic registry lookups, having this **check** here completes the function on its own. Similarly how we validate whether the topic name sent was valid or not.
Will change if this doesn't sound correct. Thoughts?

internal/topic/core.go
```diff
@@ -91,6 +92,25 @@ func (c *Core) CreateDeadLetterTopic(ctx context.Context, model *Model) error {
...
+	// validate if the topic exists
+	ok, err := c.Exists(ctx, m.Key())
+	if err != nil {
+		return err
+	}
+	if !ok {
+		return merror.New(merror.NotFound, "topic not found")
+	}
+
+	return c.repo.Create(ctx, m)
```
sunnyagg, line 111: Create or Update ?

shivasishdas: Its the same thing. Internally consul `PUT` gets called.

internal/topic/validation_test.go
```diff
@@ -46,4 +46,33 @@ func TestValidation_GetValidatedModel(t *testing.T) {
...
 	assert.Equal(t, m.Name, topic)
+	assert.Equal(t, m.NumPartitions, DefaultNumPartitions)
+}
+
+func TestValidation_GetValidatedAdminModel(t *testing.T) {
+	ctx := context.Background()
+
+	topic := "projects/test-project/topics/test-topic"
+	numPartitions := 5
+	m, err := GetValidatedTopicForAdminUpdate(ctx, &metrov1.AdminTopic{
```
sunnyagg, line 57: for starting we should just allow update of `NumPartitions` ?

internal/topic/validation.go
```diff
@@ -37,6 +37,25 @@ func GetValidatedModel(ctx context.Context, req *metrov1.Topic) (*Model, error)
...
 }
 
+// GetValidatedTopicForAdminUpdate validates an incoming proto request and returns the model
+func GetValidatedTopicForAdminUpdate(ctx context.Context, req *metrov1.AdminTopic) (*Model, error) {
+	p, t, err := ExtractTopicMetaAndValidateForCreate(ctx, req.GetName())
+	if err != nil {
+		return nil, merror.Newf(merror.InvalidArgument, "Invalid [topics] name: (name=%s)", req.Name)
+	}
+	m := &Model{}
+	m.Name = req.GetName()
```
sunnyagg, line 47: this shouldn't be allowed for update ?

internal/topic/validation_test.go
shivasishdas: Currently, only `NumPartitions` gets updated. `Name` is only used to identify the topic. Kept `Labels` for compatibility with the existing `Topic` model.

internal/topic/validation.go
shivasishdas: Lookup on topic `name` will be done using this. If a wrong/different name is sent, say for the update, then this lookup would fail.

sunnyagg (approved):

Issue #125, opened 2021-05-24
On scaling in workers, subscriptions do not get rescheduled


shivasishdas:
`Scenario:`

- Scale-up worker replica set with 2: `w1` and w2`
`- Create a subscription on a topic with 5 partitions : `tp1`,`tp2`,`tp3`,`tp4`,`tp5`
- Now `tp1`,`tp2` get scheduled on `w1` and `tp3`,`tp4`,`tp5` get scheduled on `w2`.
- Scale down the worker replica set to 1. 
-  `w2` gets removed along with `tp3`,`tp4`,`tp5`.

`Issue:`
- `tp3`,`tp4`,`tp5` worker hanlders never get re-scheduled.

PR #129, opened 2021-05-25
handles rescheduling due to node failure

sunnyagg:
Issue #125 
For node updates, if we find a binding which was associated with a deleted node, we should reschedule it (than deleting it)


shivasishdas (commented):
LGTM

shivasishdas (approved):

PR #131, opened 2021-05-31
support basic auth on push endpoint

shivasishdas:
- Accepting basic auth creds during subscription creation. Sample payload below

```
{
    "topic": "projects/perf10/topics/topic123",
    "push_config": {
        "push_endpoint": "https://webhook.site/1b154a15-421c-4a5a-8e62",
        "attributes": {
            "username": "admin",
            "password": "supersecret"
        }
    }
}
```

- The username and password will be persisted as part of the subscription model with the password kept as `base64` encoded.

### DIFF
--- a/internal/common/utils.go
+++ b/internal/common/utils.go
@@ -1,0 +1,14 @@
+package common
+
+import "encoding/base64"
+
+// Encode given a string, returns a base64 encoded version of it
+func Encode(input string) string {
+	return base64.StdEncoding.EncodeToString([]byte(input))
+}
+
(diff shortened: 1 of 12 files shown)


internal/common/utils.go
```diff
@@ -0,0 +1,14 @@
+package common
```
sunnyagg, line 1: move to pkg ?

shivasishdas: done

sunnyagg (approved):

PR #134, opened 2021-06-02
App/Admin auth middleware impl

shivasishdas:
**Implements two specific auth middlewares**
- `Admin Auth`: These creds would be read from the config injected via env.
- `App Auth`: These are project-specific access key pairs generated via the Admin APIs. Ideally, every project would get a unique access key pair.

### DIFF
--- a/service/web/service.go
+++ b/service/web/service.go
@@ -3,6 +3,7 @@ package web
 import (
 	"context"
 	"net/http"
+	"os"
 
 	"github.com/razorpay/metro/internal/interceptors"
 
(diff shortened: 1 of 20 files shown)


service/web/service.go
```diff
@@ -158,5 +164,14 @@ func (svc *Service) Stop() error {
 }
 
 func getInterceptors() []grpc.UnaryServerInterceptor {
-	return []grpc.UnaryServerInterceptor{}
+	// skip auth from test mode executions
```
sunnyagg, line 167: why ?

service/web/subscriberserver.go
```diff
@@ -4,6 +4,9 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/razorpay/metro/internal/interceptors"
+	"github.com/razorpay/metro/internal/project"
+
```
sunnyagg, line 9: nit; remove extra lines

shivasishdas:
Match the key:projectID and the route:projectID in middleware.

service/web/service.go
```diff
@@ -158,5 +164,14 @@ func (svc *Service) Stop() error {
...
+	if os.Getenv("EXEC_MODE") == "test" {
```
shivasishdas, line 168: Figure out a better way to handle this

PR #136, opened 2021-06-03
set topic replication factor to the number of brokers

shivasishdas:
Refer: https://kafka.apache.org/documentation/#basic_ops_add_topic

### DIFF
--- a/pkg/messagebroker/kafka.go
+++ b/pkg/messagebroker/kafka.go
@@ -218,7 +218,7 @@ func (k *KafkaBroker) CreateTopic(ctx context.Context, request CreateTopicReques
 	ts := kafkapkg.TopicSpecification{
 		Topic:             tp,
 		NumPartitions:     request.NumPartitions,
-		ReplicationFactor: len(k.Config.Brokers),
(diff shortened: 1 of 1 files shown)


pkg/messagebroker/kafka.go
```diff
@@ -218,7 +218,7 @@ func (k *KafkaBroker) CreateTopic(ctx context.Context, request CreateTopicReques
 	ts := kafkapkg.TopicSpecification{
 		Topic:             tp,
 		NumPartitions:     request.NumPartitions,
-		ReplicationFactor: 1,
+		ReplicationFactor: len(k.Config.Brokers),
```
sunnyagg, line 221: high value will have impact on publish latency as data will need time to replicate in sync. we should start with atleast 50% number 

i.e. (len(k.Config.Brokers)+1)/2

sunnyagg (approved):

pkg/messagebroker/kafka.go
sunnyagg: Plus more replication will add to cost as well in terms of storage and bytes over network

PR #151, opened 2021-06-14
better health checks

shivasishdas:
- Implements a health checks over Broker and Registry

### DIFF
--- a/internal/health/core.go
+++ b/internal/health/core.go
@@ -2,18 +2,15 @@ package health
 
 import (
 	"context"
-	"fmt"
 	"sync"
 
 	"github.com/razorpay/metro/pkg/logger"
-	"github.com/razorpay/metro/pkg/messagebroker"
-	"github.com/razorpay/metro/pkg/registry"
 )
 
 // Checker interface for health
 type Checker interface {
 	name() string
-	checkHealth() (bool, error)
+	checkHealth(ctx context.Context) (bool, error)
 }
 
 // Core holds business logic and/or orchestrator of other things in the package.
@@ -30,7 +27,7 @@ func NewCore(checkers ...Checker) (*Core, error) {
 
 // IsHealthy checks if the app has been marked unhealthy. If so it'll return false. Otherwise, it'll check the application
 // health and return a boolean value based on the health check result.
-func (c *Core) IsHealthy() bool {
+func (c *Core) IsHealthy(ctx context.Context) bool {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 	// Server has been marked as unhealthy. No point in doing a health check.
@@ -46,7 +43,7 @@ func (c *Core) IsHealthy() bool {
 		if checker == nil {
 			continue
 		}
-		isHealthy, err := checker.checkHealth()
+		isHealthy, err := checker.checkHealth(ctx)
 		if !isHealthy || err != nil {
 			logger.Ctx(context.TODO()).Errorw("health check failed", "service_to_check", checker.name(), "error", err)
 			return false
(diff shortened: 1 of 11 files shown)


internal/health/core.go
```diff
@@ -43,8 +47,8 @@ func (c *Core) IsHealthy() bool {
 			continue
 		}
 		isHealthy, err := checker.checkHealth()
-		if !isHealthy {
-			logger.Ctx(context.TODO()).Errorw("health check failed", "msg", err)
+		if !isHealthy || err != nil {
```
sunnyagg, line 50: break the err case handling in separate if block
```suggestion
		if err != nil {
		log err.Error
        handle err based on err types and decide to return true/false
		}
		
		return isHealthy
```

internal/health/core.go
```diff
@@ -60,13 +64,38 @@ func (c *Core) MarkUnhealthy() {
 	c.isMarkedUnhealthy = true
 }
 
-type dbHealthChecker struct{}
+type registryHealthChecker struct {
```
sunnyagg, line 67: move to separate files ?

pkg/messagebroker/kafka.go
```diff
@@ -124,6 +126,7 @@ func newKafkaProducerClient(ctx context.Context, bConfig *BrokerConfig, options
 	logger.Ctx(ctx).Infow("kafka producer: initialized")
 
 	return &KafkaBroker{
+		Ctx:      ctx,
```
sunnyagg, line 129: pass this with function call ?

pkg/messagebroker/messagebroker.go
```diff
@@ -19,6 +19,9 @@ type Admin interface {
 
 	// adds partitions to an existing topic
 	AddTopicPartitions(context.Context, AddTopicPartitionRequest) (*AddTopicPartitionResponse, error)
+
+	// checks health of underlying broker
+	IsHealthy() (bool, error)
```
sunnyagg, line 24: add ctx

pkg/registry/consul.go
```diff
@@ -301,3 +301,9 @@ func (c *ConsulClient) DeleteTree(key string) error {
 	}
 	return nil
 }
+
+// IsAlive checks the health of the consul
+func (c *ConsulClient) IsAlive() (bool, error) {
```
sunnyagg, line 306: add ctx here too, will add jaeger instrumentation on this after the tracing pr is merged

internal/health/core.go
shivasishdas: If you look at the full code snippet these checks are running inside a loop. So having `!isHealthy || err != nil` is a fail-fast implementation. In the case of healthy checks, we just let the loop complete and return from the end of the function.

internal/health/core.go
shivasishdas: done

pkg/messagebroker/kafka.go
```diff
@@ -639,3 +639,29 @@ func (k *KafkaBroker) AddTopicPartitions(ctx context.Context, request AddTopicPa
 		Response: resp,
 	}, nil
 }
+
+// IsHealthy checks the health of the kafka
+func (k *KafkaBroker) IsHealthy(ctx context.Context) (bool, error) {
+	doneCh := make(chan healthResponse)
+
+	go func() {
```
sunnyagg, line 647: can we create a context with Deadline of 1 second and pass that on to this method, if library handles the timeout defined based on ctx, we can avoid this

sunnyagg (approved):

pkg/messagebroker/kafka.go
shivasishdas: Had thought of this. But the comment on top of the function impl warns the timeout may not be honored. Hence did it this way. Maybe I'm interpreting it wrong.

```
// Note on cancellation: Although the underlying C function respects the
// timeout, it currently cannot be manually cancelled. That means manually
// cancelling the context will block until the C function call returns.
```

https://github.com/confluentinc/confluent-kafka-go/blob/13ae1158cd0e098a1093c52d77d556a5c733e242/kafka/adminapi.go#L430

sunnyagg: i think it will handle the timeout value but we can't cancel context at our end, so we can use context.WithDeadline but we can't use context.WithCancel

shivasishdas: Made the change

PR #170, opened 2021-06-16
prod configs

shivasishdas:


config/prod.toml
```diff
@@ -0,0 +1,69 @@
+[app]
+    appEnv                = "prod"
+    serviceName           = "metro"
+    shutdownTimeout       = 2
+    shutdownDelay         = 2
+    gitCommitHash         = "nil"
+[tracing]
+    serviceName           = "metro"
+    logSpans              = false
+    localAgentHostPort    = "localhost:6831"
```
sunnyagg, line 10: changed to host, port in latest code, should we leave this to default.toml value if not defined

config/prod.toml
```diff
@@ -0,0 +1,69 @@
...
+    disabled              = true
```
sunnyagg, line 11: false

config/prod.toml
shivasishdas: changed. Took latest master.

sunnyagg (approved):

PR #175, opened 2021-06-17
Metro Docs: fix for base url

sunnyagg:
- fixes docs path prefix (removes extra /metro)


sunnyagg:
Adding docs in separate PR, for now just fixing path

shivasishdas (approved):

PR #177, opened 2021-06-18
set http timeout using ack_deadline_seconds

shivasishdas:
- Using subscription `ack_deadline_seconds` to set http timeout Refer [line](https://github.com/razorpay/metro-proto/blob/d3941c9042bb0feda28baf7ff2376718cee3deba/metro/proto/v1/spec.proto#L386)

- Setting default HTTP timeout to `10sec` in case `ack_deadline_seconds`  is not sent during the subscription creation. Refer [line](https://github.com/razorpay/metro-proto/blob/d3941c9042bb0feda28baf7ff2376718cee3deba/metro/proto/v1/spec.proto#L384)

### DIFF
--- a/service/worker/stream.go
+++ b/service/worker/stream.go
@@ -23,6 +23,7 @@ type PushStream struct {
 	ctx              context.Context
 	cancelFunc       func()
 	nodeID           string
+	subscription     *subscription.Model
 	subscriptionName string
 	subscriptionCore subscription.ICore
 	subscriberCore   subscriber.ICore
@@ -38,7 +39,7 @@ func (ps *PushStream) Start() error {
 	var (
 		err error
 		// init these channels and pass to subscriber
-		// the lifecycle of these channels should be maintain by the user
+		// the lifecycle of these channels should be maintained by the user
 		subscriberRequestCh = make(chan *subscriber.PullRequest)
 		subscriberAckCh     = make(chan *subscriber.AckMessage)
 		subscriberModAckCh  = make(chan *subscriber.ModAckMessage)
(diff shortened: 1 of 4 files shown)


service/worker/stream.go
```diff
@@ -189,15 +183,31 @@ func (ps *PushStream) ack(ctx context.Context, message *metrov1.ReceivedMessage)
...
+		// make sure to convert sec to milli-sec
+		config.ConnectTimeoutMS = int(subModel.AckDeadlineSec) * 1e3
+	}
+	httpclient := NewHTTPClientWithConfig(config)
+
 	return &PushStream{
 		ctx:              pushCtx,
 		cancelFunc:       cancelFunc,
 		nodeID:           nodeID,
+		subscription:     subModel,
```
sunnyagg, line 205: if we are passing model, we can remove the subcriptionName attribute below

shivasishdas: Was keeping it to keep the file diff small :) Anyways did the change now.

service/worker/stream.go
```diff
@@ -38,23 +38,16 @@ func (ps *PushStream) Start() error {
...
 		// init these channels and pass to subscriber
-		// the lifecycle of these channels should be maintain by the user
+		// the lifecycle of these channels should be maintained by the user
 		subscriberRequestCh = make(chan *subscriber.PullRequest)
 		subscriberAckCh     = make(chan *subscriber.AckMessage)
 		subscriberModAckCh  = make(chan *subscriber.ModAckMessage)
 	)
 
-	ps.subs, err = ps.subscriberCore.NewSubscriber(ps.ctx, ps.nodeID, ps.subscriptionName, 100, 50, 0,
+	ps.subs, err = ps.subscriberCore.NewSubscriber(ps.ctx, ps.nodeID, ps.subscription.ExtractedSubscriptionName, 100, 50, 0,
```
sunnyagg, line 47: why extracted subscription name, full name will give project id as well. that will be easy to debug ?

shivasishdas: Sure. So use `ps.subscription.Key()` instead everywhere?

sunnyagg: Key will be our internal path 
/subscriptions/{projectid}/{extracted sub name}

we should log the public id i think

shivasishdas: logging `subscription.name` now

sunnyagg (approved):

PR #180, opened 2021-06-22
use different endpoints for readiness and liveness

shivasishdas:
-Proto changes https://github.com/razorpay/metro-proto/pull/21/files

### DIFF
--- a/internal/health/server.go
+++ b/internal/health/server.go
@@ -15,6 +15,7 @@ type Server struct {
 	core *Core
 }
 
+// ReadinessCheck checks only for the responsiveness of the service
 func (h *Server) ReadinessCheck(ctx context.Context, _ *emptypb.Empty) (*metrov1.HealthCheckResponse, error) {
 	if !h.core.IsHealthy(ctx) {
 		logger.Ctx(ctx).Debugw("metro health check", "status", "Unhealthy")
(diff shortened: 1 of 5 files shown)


sunnyagg (approved):

internal/health/server.go
```diff
@@ -7,27 +7,49 @@ import (
...
+	}
+
+	logger.Ctx(ctx).Debugw("metro readiness check", "status", "Healthy")
+	return &metrov1.StatusCheckResponse{
+		ServingStatus: metrov1.StatusCheckResponse_SERVING_STATUS_SERVING,
+	}, nil
+}
+
+// LivenessCheck checks responsiveness of all the dependant resources that the service is using
+func (h *Server) LivenessCheck(ctx context.Context, _ *emptypb.Empty) (*metrov1.StatusCheckResponse, error) {
```
sunnyagg, line 32: by default all grpc apis are being tracked under opentracing, we can remove liveliness prove from there as k8s checks that every sec and adds to lot of volume

shivasishdas: ah yes! let me filter these out.

shivasishdas: anyways will make this change in a different PR. Need to figure out the middleware interceptor.

sunnyagg: Sure, Works for now

shivasishdas: Tracking here https://github.com/razorpay/metro/issues/183

PR #184, opened 2021-06-24
adds unit tests for consul client

sunnyagg:
Tests consul client by mocking consul http api

### DIFF
--- a/pkg/registry/consul_test.go
+++ b/pkg/registry/consul_test.go
@@ -3,7 +3,15 @@
 package registry
 
 import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/google/uuid"
+	"github.com/hashicorp/consul/api"
+	"net/http"
+	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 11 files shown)


pkg/registry/consul_test.go
```diff
@@ -3,15 +3,605 @@
...
 	config := ConsulConfig{}
 	c1, err := NewConsulClient(&config)
-	t.Log(err)
 	assert.NotNil(t, c1)
 	assert.Nil(t, err)
 }
+
+func TestRegisterSuccess(t *testing.T) {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/v1/session/create", func(res http.ResponseWriter, req *http.Request) {
```
shivasishdas, line 28: nit: Seeing a lot of duplicate code. Maybe init two common HTTP servers (one for success and one for failure) with all routes and re-use. Use [TestMain](https://golang.org/pkg/testing/#hdr-Main)?

shivasishdas (changes requested):

pkg/registry/consul_test.go
sunnyagg: issue with TestMain is that that is not per package, had explored using init() in file but there is no reverse for init 
Per test made sense as generally we define mock data as part of tests
if we create S and F as 2 separate things failures can have multiple types, creating combinations of failures will eventually mean defining those many combinations :(

shivasishdas (approved):

PR #188, opened 2021-07-02
admin auth fixes

shivasishdas:
- With the introduction of credentials, worker calls were not getting authenticated. Have set admin creds in ctx() during init() to fix this.

- Another way to bypass auth for workers can be by reading APP_COMPONENT from the env. Thougts?


shivasishdas:
No longer needed.

PR #191, opened 2021-07-04
encrypt creds

shivasishdas:


cmd/service/metro/metro.go
```diff
@@ -61,6 +62,11 @@ func Init(ctx context.Context, env string, componentName string) {
 		log.Fatal("admin credentials missing")
 	}
 
+	if appConfig.Encryption.Key == "" {
+		log.Fatal("encryption key missing")
+	}
+	encryption.RegisterEncryptionKey(appConfig.Encryption.Key)
```
sunnyagg, line 68: should we create a singlton instance and pass the whole encryption config than just key

sunnyagg (approved):

sunnyagg:
what about webhook creds ?

shivasishdas:
> what about webhook creds?

While creating a subscription we store the credentials using `credentials.NewCredential()`. So the encryption will get applied there as well.
https://github.com/razorpay/metro/blob/a4282fa6bd873697f1b9983ae712040229ec9a05/internal/subscription/validation.go#L71

Issue #193, opened 2021-07-07
Validate project access in middleware


shivasishdas:
The below enhancement allows access matched HTTP pattern in gRPC gateway. Need to leverage this in the middleware to enforce project <> project-credentials access.

https://grpc-ecosystem.github.io/grpc-gateway/docs/operations/annotated_context/

Currently, since the project entity is not available in the middleware, as a workaround we are using `credentials.IsAuthorized()` from within `topic` and `subscription` validations to check and fail access.

https://github.com/razorpay/metro/blob/a4282fa6bd873697f1b9983ae712040229ec9a05/internal/topic/validation.go#L101
https://github.com/razorpay/metro/blob/a4282fa6bd873697f1b9983ae712040229ec9a05/internal/subscription/validation.go#L151

PR #194, opened 2021-07-07
enforce projectID from middleware

shivasishdas:
Closes https://github.com/razorpay/metro/issues/193

Have filed a bug as well on implementation: https://github.com/grpc-ecosystem/grpc-gateway/issues/2222

### DIFF
--- a/internal/interceptors/auth.go
+++ b/internal/interceptors/auth.go
@@ -24,11 +24,15 @@ func UnaryServerAuthInterceptor(authFunc grpc_auth.AuthFunc) grpc.UnaryServerInt
 	return grpc_auth.UnaryServerInterceptor(authFunc)
 }
 
-func getUserPassword(ctx context.Context) (user string, password []byte, err error) {
+func getUserPasswordProjectID(ctx context.Context) (user string, password []byte, projectID string, err error) {
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
 		err = status.Error(codes.Unauthenticated, "could not parse incoming context")
 		return
+	}
+
+	if val := md.Get("uri"); val != nil && len(val) > 0 {
+		projectID = extractProjectIDFromURI(val[0])
 	}
 
 	headers := md.Get(authorizationHeaderKey)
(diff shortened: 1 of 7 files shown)


internal/interceptors/auth.go
```diff
@@ -24,13 +24,17 @@ func UnaryServerAuthInterceptor(authFunc grpc_auth.AuthFunc) grpc.UnaryServerInt
...
 
-func getUserPassword(ctx context.Context) (user string, password []byte, err error) {
+func getUserPasswordProjectID(ctx context.Context) (user string, password []byte, projectID string, err error) {
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
 		err = status.Error(codes.Unauthenticated, "could not parse incoming context")
 		return
 	}
 
+	if val := md.Get("uri"); val != nil && len(val) > 0 {
```
sunnyagg, line 34: add comments, is val a slice ?

internal/interceptors/auth.go
```diff
@@ -68,9 +72,25 @@ func secureCompare(expected, actual string) bool {
 	return false
 }
 
+// extractProjectIDFromURI : This function extracts the projectID from the HTTP request URI.
+// Example1: extractProjectIDFromURI("/v1/projects/project1/topics/t123") = project1
+// Example2: extractProjectIDFromURI("/v1/projects/project7/subscriptions/s123") = project7
+// Example3: extractProjectIDFromURI("/v1/admin/topic/t987") = ""
```
sunnyagg, line 78: will this middleware be called for admin routes ? example 3 doesn't seem to be valid. this middleware should be only for client auth. admin auth can be a different middleware

internal/interceptors/auth.go
```diff
@@ -68,9 +72,25 @@ func secureCompare(expected, actual string) bool {
...
+func extractProjectIDFromURI(uri string) string {
+	if uri == "" {
+		return ""
+	}
+
+	parts := strings.Split(uri, "/")
```
sunnyagg, line 84: should we use some regex here
v1/projects/{id}/{anythingelse}

internal/interceptors/auth.go
shivasishdas: Middlewares are different. Both of them are using a common function to extract `getUserPasswordProjectID()` from the `authorization` header. This is needed for both the auth types.

internal/interceptors/auth.go
shivasishdas: added

sunnyagg (changes requested):
Add tests, coverage is going down

shivasishdas:
> Add tests, coverage is going down

Adding tests for AppAuth and AdminAuth

sunnyagg (approved):

PR #201, opened 2021-07-16
Adds unit tests 

sunnyagg:
- Adds Unit tests for subscription manager

### DIFF
--- a/internal/stream/stream.go
+++ b/internal/stream/stream.go
@@ -44,7 +44,7 @@ func (ps *PushStream) Start() error {
 		subscriberModAckCh  = make(chan *subscriber.ModAckMessage)
 	)
 
-	// we pass a new context to subscriber because if the subscriber gets a child context of the steam context. there
+	// we pass a new context to subscriber because if the subscriber gets a child context of the stream context. there
 	// is a race condition that subscriber exits before the stream is stopped. this causes issues as there are no
 	// subscribers listening to the requests send by stream
 	subscriberCtx := context.Background()
--- a/internal/stream/stream.go
+++ b/internal/stream/stream.go
@@ -44,7 +44,11 @@ func (ps *PushStream) Start() error {
 		subscriberModAckCh  = make(chan *subscriber.ModAckMessage)
 	)
 
-	ps.subs, err = ps.subscriberCore.NewSubscriber(ps.ctx, ps.nodeID, ps.subscription, 100, 50, 0,
+	// we pass a new context to subscriber because if the subscriber gets a child context of the steam context. there
+	// is a race condition that subscriber exits before the stream is stopped. this causes issues as there are no
+	// subscribers listening to the requests send by stream
+	subscriberCtx := context.Background()
+	ps.subs, err = ps.subscriberCore.NewSubscriber(subscriberCtx, ps.nodeID, ps.subscription, 100, 50, 0,
 		subscriberRequestCh, subscriberAckCh, subscriberModAckCh)
 	if err != nil {
 		logger.Ctx(ps.ctx).Errorw("worker: error creating subscriber", "subscription", ps.subscription.Name, "error", err.Error())
--- a/internal/subscriber/subscriber.go
+++ b/internal/subscriber/subscriber.go
@@ -27,8 +27,8 @@ type ISubscriber interface {
 	GetSubscription() string
 	// not exporting acknowledge() and  modifyAckDeadline() intentionally so that
 	// all operations happen over the channel
-	acknowledge(ctx context.Context, req *AckMessage)
-	modifyAckDeadline(ctx context.Context, req *ModAckMessage)
+	//acknowledge(ctx context.Context, req *AckMessage)
+	//modifyAckDeadline(ctx context.Context, req *ModAckMessage)
 	// the grpc proto is used here as well, to optimise for serialization
 	// and deserialisation, a little unclean but optimal
 	// TODO: figure a better way out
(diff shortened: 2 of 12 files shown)


internal/tasks/leadertask.go
```diff
@@ -0,0 +1,146 @@
...
+	return err
+}
+
+func (sm *LeaderTask) runLeaderElection(ctx context.Context, sessionID string) error {
+	// Init Leader Election
+	candidate, err := leaderelection.New(
+		sm.id,
+		sessionID,
+		leaderelection.Config{
+			LockPath: common.GetBasePrefix() + "leader/election",
```
razorao, line 120: should "leader/election" be a constant ?

internal/tasks/leadertask_test.go
```diff
@@ -0,0 +1,125 @@
...
+	task, err := NewLeaderTask(workerID, registryMock, nodeCoreMock, taskMock)
+	assert.Nil(t, err)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Mock responses to registry
+	registryMock.EXPECT().Register(gomock.Any(), "metro/metro-worker", 30*time.Second).Return("id", nil).AnyTimes()
+	registryMock.EXPECT().RenewPeriodic(gomock.Any(), "id", 30*time.Second, gomock.Any()).Return(nil).AnyTimes()
+	registryMock.EXPECT().Release(gomock.Any(), "id", common.GetBasePrefix()+"leader/election", workerID).Return(true)
```
razorao, line 76: "leader/election" constant?

internal/stream/stream.go
```diff
@@ -44,7 +44,11 @@ func (ps *PushStream) Start() error {
 		subscriberModAckCh  = make(chan *subscriber.ModAckMessage)
 	)
 
-	ps.subs, err = ps.subscriberCore.NewSubscriber(ps.ctx, ps.nodeID, ps.subscription, 100, 50, 0,
+	// we pass a new context to subscriber because if the subscriber gets a child context of the steam context. there
```
shivasishdas, line 47:
```suggestion
	// we pass a new context to subscriber because if the subscriber gets a child context of the stream context. there
```

internal/stream/stream.go
```diff
@@ -44,7 +44,11 @@ func (ps *PushStream) Start() error {
...
+	// is a race condition that subscriber exits before the stream is stopped. this causes issues as there are no
+	// subscribers listening to the requests send by stream
+	subscriberCtx := context.Background()
```
shivasishdas, line 50: `subscriber exits before the stream is stopped`: cant we handle this using the `ctx.Timeout()` to avoid a blocking call in case the subscriber channels get closed?

internal/subscriber/subscriber.go
```diff
@@ -27,8 +27,8 @@ type ISubscriber interface {
 	GetSubscription() string
 	// not exporting acknowledge() and  modifyAckDeadline() intentionally so that
 	// all operations happen over the channel
-	acknowledge(ctx context.Context, req *AckMessage)
-	modifyAckDeadline(ctx context.Context, req *ModAckMessage)
```
shivasishdas, line 31: ?

internal/tasks/leadertask.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, option := range options {
+		option(leaderTask)
+	}
+
+	return leaderTask, nil
+}
+
+func defaultOptions() []Option {
+	return []Option{
+		WithTTL(30 * time.Second),
```
shivasishdas, line 55: nit: can create constants for these values as well

internal/tasks/schedulertask_test.go
```diff
@@ -1,81 +1,128 @@
+// +build unit
```
shivasishdas, line 1: Additionally, we run tests using the `-race` flag. https://golang.org/doc/articles/race_detector#Usage

shivasishdas (changes requested):

internal/stream/stream.go
sunnyagg: ctx.timeout will just wait until timeout, it wont control the exit behaivour of the subscriber

shivasishdas: Not here. Was suggesting having this at places where we read off like `<-subscriber.channels()`. A timeout will ensure that the request is not blocked forever?

internal/subscriber/subscriber.go
sunnyagg: Since these methods are not exported, there was some issue with mock generated. generated mock wasn't assignable to the interface. After commenting this it worked. mostly a issue in golang

internal/tasks/schedulertask_test.go
sunnyagg: had tried running it with count 100-1000, will check this, possibly will make sense to all all tests

internal/stream/stream.go
sunnyagg: that wont work too in this particular case as a select case is already executing, in this particular case our default block was executing, but deriving from the thought process, we shouldn't have a default block. we should move the default processing to a case statement. that way if the case statement is blocked. it will still allow the ctx.Done to execute

internal/subscriber/subscriber.go
sunnyagg: https://github.com/razorpay/metro/pull/201/files#diff-1d4ef1c48c1b1aff0ea88f681a7e7c463724dd56ce27f6bbd2d17f804c11a809R188

This fails to handle the type with unexported method defined in the interface

shivasishdas (approved):

PR #202, opened 2021-07-16
cleaning up logs a bit

shivasishdas:
Will do it at other places as well once the refactor PRs are merged.


internal/subscriber/subscriber.go
```diff
@@ -541,9 +554,18 @@ func (s *Subscriber) logInMemoryStats() {
 		}
 		st[tp.String()] = total
 	}
+	logger.Ctx(s.ctx).Infow("subscriber: in-memory stats", s.getLogFields(), "stats", st)
+}
 
-	logger.Ctx(s.ctx).Infow("subscriber: in-memory stats", "stats", st,
-		"topic", s.topic, "subscription", s.subscription, "subscriberId", s.subscriberID)
+// returns a map of common fields to be logged
+func (s *Subscriber) getLogFields() map[string]interface{} {
```
sunnyagg, line 561: can we set them in context ?

shivasishdas: Had tried this. Values set in `ctx()` were not getting goroutine scoped. eg: `msgID` added to `ctx` of one request as getting used up in consecutive requests as well. Maybe I was doing it wrong. Looking for implementation suggestions using this.

internal/credentials/model.go
```diff
@@ -59,7 +59,7 @@ func (m *Model) Prefix() string {
 }
 
 // NewCredential returns a new credential model
-func NewCredential(username, password string) ICredentials {
+func NewCredential(username, password string) *Model {
```
sunnyagg, line 62: IModel ?

shivasishdas: Serde was failing with the interface previously hence changed it to the struct{}.

sunnyagg (approved):

PR #203, opened 2021-07-25
Edit subscription worker

razorao:
--Modified the common base model and interface to add version setters and getters

    --Moved the implementation of consul PUT API from KV to transaction based to get the modify index on edits

    --Modified common repo methods to set and return version ids

    --Modified node,nodebinding and subscription repo to set the version id in models in get calls

    --Added subscription version in nodebinding model

    --Modified scheduler to set the subscription version id when creating nodebindings

    --Modified the refresh bindings method to delete nodebindings on stale subscriptions
       and create new ones for updated subscriptions

    --Modified registry interface to incorporate version id on read and write operations

    --Modified the consul registry implementation to provide the modifyindex as versionid

    --Fixed linting errors

    --Fixed broken unit tests

### DIFF
--- a/internal/common/model.go
+++ b/internal/common/model.go
@@ -1,6 +1,13 @@
 package common
 
-import "github.com/razorpay/metro/internal/app"
+import (
+	"fmt"
+
+	"github.com/razorpay/metro/internal/app"
+)
+
+// Version of key value pair provided by the registry. Version changes on every update
+type VersionID string
 
 // IModel interface which all models should implement
 type IModel interface {
@@ -8,10 +15,32 @@ type IModel interface {
 	Key() string
 	// Prefix returns the key prefix used by the module
 	Prefix() string
+	// Returns the version of the saved model in the registry. Version changes for every update.
+	GetVersionID() (string, error)
+	//Sets the version of the saved model from the registry. Never to be used by business logic.
+	SetVersionID(vid string)
 }
 
 // BaseModel implements basic model functionality
-type BaseModel struct{}
+type BaseModel struct {
+	versionID    string
+	isVersionSet bool
+}
+
+//Base implementation of set version id
+func (b *BaseModel) SetVersionID(vid string) {
+	b.isVersionSet = true
+	b.versionID = vid
+}
+
+//Base implementation of get version id. errors out if called without setting the version id.
+func (b *BaseModel) GetVersionID() (string, error) {
+	if !b.isVersionSet {
+		err := fmt.Errorf("calling getVersionID without setting it")
+		return "", err
+	}
+	return b.versionID, nil
+}
 
 // GetBasePrefix returns the prefix
 func GetBasePrefix() string {
--- a/internal/common/model.go
+++ b/internal/common/model.go
@@ -1,9 +1,11 @@
 package common
 
 import (
-	"fmt"
-
 	"github.com/razorpay/metro/internal/app"
+)
+
+const (
+	defaultVersion string = "0"
 )
 
 // IModel interface which all models should implement
@@ -13,30 +15,27 @@ type IModel interface {
 	// Prefix returns the key prefix used by the module
 	Prefix() string
 	// Returns the version of the saved model in the registry. Version changes for every update.
-	GetVersionID() (string, error)
+	GetVersion() string
 	//Sets the version of the saved model from the registry. Never to be used by business logic.
-	SetVersionID(vid string)
+	SetVersion(version string)
 }
 
 // BaseModel implements basic model functionality
 type BaseModel struct {
-	versionID    string
-	isVersionSet bool
+	version *string
 }
 
-//SetVersionID Base implementation of set version id
-func (b *BaseModel) SetVersionID(vid string) {
-	b.isVersionSet = true
-	b.versionID = vid
+//SetVersion Base implementation of set version
+func (b *BaseModel) SetVersion(version string) {
+	b.version = &version
 }
 
-//GetVersionID Base implementation of get version id. errors out if called without setting the version id.
-func (b *BaseModel) GetVersionID() (string, error) {
-	if !b.isVersionSet {
-		err := fmt.Errorf("calling getVersionID without setting it")
-		return "", err
+//GetVersion Base implementation of get version. Default version is "0".
+func (b *BaseModel) GetVersion() string {
+	if b.version == nil {
+		return defaultVersion
 	}
-	return b.versionID, nil
+	return *b.version
 }
 
 // GetBasePrefix returns the prefix
--- a/internal/common/model.go
+++ b/internal/common/model.go
@@ -6,9 +6,6 @@ import (
 	"github.com/razorpay/metro/internal/app"
 )
 
-// Version of key value pair provided by the registry. Version changes on every update
-type VersionID string
-
 // IModel interface which all models should implement
 type IModel interface {
 	// Key returns the key to store the model against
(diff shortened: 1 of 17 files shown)


pkg/registry/consul.go
```diff
@@ -216,15 +218,23 @@ func (c *ConsulClient) Put(ctx context.Context, key string, value []byte) error
...
 	}()
 
-	_, err := c.client.KV().Put(&api.KVPair{
-		Key:   key,
-		Value: value,
-	}, nil)
-	return err
+	// Using transaction because a simple put does not fetch the modifyindex
+	putTxn := api.TxnOps{&api.TxnOp{KV: &api.KVTxnOp{Verb: api.KVSet, Key: key, Value: value}}}
+	ok, resp, _, err := c.client.Txn().Txn(putTxn, nil)
```
shivasishdas, line 223: check `err!=nil {}` as well

pkg/registry/consul.go
```diff
@@ -216,15 +218,23 @@ func (c *ConsulClient) Put(ctx context.Context, key string, value []byte) error
...
+	if !ok {
+		if err != nil {
+			return "", err
+		}
+		// There must be exactly 1 error
+		err = fmt.Errorf(resp.Errors[0].What)
+		return "", err
+	}
+
+	return strconv.FormatUint(resp.Results[0].KV.ModifyIndex, 10), err
```
shivasishdas, line 233: add safe checks before accessing nested field to avoid panics.
`len(resp.Results)>0 && resp.Results[0].KV !=nil`

pkg/registry/consul.go
```diff
@@ -216,15 +218,23 @@ func (c *ConsulClient) Put(ctx context.Context, key string, value []byte) error
...
-	return err
+	// Using transaction because a simple put does not fetch the modifyindex
+	putTxn := api.TxnOps{&api.TxnOp{KV: &api.KVTxnOp{Verb: api.KVSet, Key: key, Value: value}}}
+	ok, resp, _, err := c.client.Txn().Txn(putTxn, nil)
+	if !ok {
+		if err != nil {
+			return "", err
+		}
+		// There must be exactly 1 error
+		err = fmt.Errorf(resp.Errors[0].What)
```
shivasishdas, line 229: add safe checks just to be sure as commented below.

pkg/registry/registry.go
```diff
@@ -46,10 +47,10 @@ type IRegistry interface {
 	Watch(ctx context.Context, wh *WatchConfig) (IWatcher, error)
 
 	// Put a key value pair
-	Put(ctx context.Context, key string, value []byte) error
+	Put(ctx context.Context, key string, value []byte) (string, error)
 
 	// Get returns a value for a key
-	Get(ctx context.Context, key string) ([]byte, error)
+	Get(ctx context.Context, key string) (Pair, error)
```
shivasishdas, line 53: pointer returns?

pkg/registry/registry.go
```diff
@@ -46,10 +47,10 @@ type IRegistry interface {
 	Watch(ctx context.Context, wh *WatchConfig) (IWatcher, error)
 
 	// Put a key value pair
```
shivasishdas, line 49: update function doc

pkg/registry/registry.go
```diff
@@ -10,11 +10,12 @@ import (
 type Pair struct {
 	Key       string
 	Value     []byte
+	VersionID string
```
shivasishdas, line 13: add a one-liner above these to explain the purpose of these fields. Maybe add that this internally maps to the `ModifyIndex` field of `consul`.

shivasishdas (changes requested):

pkg/registry/consul.go
razorao: It's being done inside `if !ok`
If err is not nil, then ok will not be true

shivasishdas: // If there is a problem making the transaction request then an error will be
// returned. Otherwise, the ok value will be true if the transaction succeeded
// or false if it was rolled back.
Was going through API doc, seems correct 👍🏼

internal/common/model.go
```diff
@@ -1,17 +1,43 @@
...
 // IModel interface which all models should implement
 type IModel interface {
 	// Key returns the key to store the model against
 	Key() string
 	// Prefix returns the key prefix used by the module
 	Prefix() string
+	// Returns the version of the saved model in the registry. Version changes for every update.
+	GetVersionID() (string, error)
+	//Sets the version of the saved model from the registry. Never to be used by business logic.
+	SetVersionID(vid string)
```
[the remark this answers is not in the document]
razorao: This would mean that setting the version id should happen in the same package where `BaseModel` is defined? But version ID has to be set in the `repo` of each domain. Am I missing something?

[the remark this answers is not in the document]
shivasishdas: This will happen inside the `Get()` of `common/repo.go`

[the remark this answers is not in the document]
razorao: Created Issue #210 to handle this

internal/common/model.go
```diff
@@ -1,17 +1,43 @@
...
+)
 
 // IModel interface which all models should implement
 type IModel interface {
 	// Key returns the key to store the model against
 	Key() string
 	// Prefix returns the key prefix used by the module
 	Prefix() string
+	// Returns the version of the saved model in the registry. Version changes for every update.
+	GetVersionID() (string, error)
```
sunnyagg, line 16: Just call them and GetVersion() and SetVersion()

internal/common/model.go
```diff
@@ -1,17 +1,43 @@
...
+
+//SetVersionID Base implementation of set version id
+func (b *BaseModel) SetVersionID(vid string) {
+	b.isVersionSet = true
+	b.versionID = vid
+}
+
+//GetVersionID Base implementation of get version id. errors out if called without setting the version id.
+func (b *BaseModel) GetVersionID() (string, error) {
+	if !b.isVersionSet {
```
sunnyagg, line 35: can we just do a nil check on version attribute instead ?

internal/tasks/schedulertask.go
```diff
@@ -203,17 +203,27 @@ func (sm *SchedulerTask) refreshNodeBindings(ctx context.Context) error {
...
 		found := false
 		for _, sub := range sm.subCache {
+			sVID, verr := sub.GetVersionID()
+			if verr != nil {
+				return verr
+			}
 			if sub.Name == nb.SubscriptionID {
-				found = true
-				break
+				if nb.SubscriptionVersionID == sVID {
```
sunnyagg, line 221: merge with if above than nesting ?

internal/common/model.go
sunnyagg: As discussed, lets do a compat test else we can assume current version to be 0 in all entities

internal/tasks/schedulertask.go
razorao: This was done on purpose so that in case where the subscription matches but the version does not match, we print a log saying the version changed. If I have to merge:
```
if (subscription.name = binding.subscriptionid) && (binding.subscriptionVersion == subscriptionVersion){
found = true
break
}
if (subscription.name = binding.subscriptionid) && (binding.subscriptionVersion != subscriptionVersion){
print log("Subscription matched but version did not match. Subscription has been updated.")
}
```
This I felt is less readable/difficult to understand

internal/common/model.go
razorao: Compat test done. Subscriptions are scheduled correctly. Have set default version as "0"

internal/scheduler/scheduler.go
```diff
@@ -39,11 +39,16 @@ func (s *Scheduler) Schedule(subscription *subscription.Model, nbs []*nodebindin
 	if err != nil {
 		return nil, err
 	}
+	subVersion := subscription.GetVersion()
+	if err != nil {
```
sunnyagg, line 43: remove this

sunnyagg (approved):

shivasishdas (approved):

PR #204, opened 2021-07-25
Edit subscription api

razorao:
--Updated the metro-proto submodule with the API contract

	--Wrote the API implementation (server and core)

	--Added validation methods to validate the API request

	--Deleted the unwanted auth check in validation

	--Added new file proto.go for translating internal models to external proto

        --Added package patch for struct field patching

        --Moved subscription patching to the new package

        --Lint checks corrected

Request:
PATCH /v1/projects/project123/subscriptions/subscription1234
Content-Type: application/json
```
{
    "subscription": {
        "topic": "projects/project123/topics/topic123",
        "push_config": {
            "push_endpoint": "http://www.ikypkkijk099pooooonky.com",
            "attributes": {
                "username": "shashank",
                "password": "password"
            }
        },
        "ackDeadlineSeconds": 15
    },
    "update_mask": "pushConfig,ackDeadlineSeconds"
}
```

Response:
200 Ok
```
{
    "name": "projects/project123/subscriptions/subscription1234",
    "topic": "projects/project123/topics/topic123",
    "pushConfig": {
        "pushEndpoint": "http://www.ikypkkijk099pooooonky.com",
        "attributes": {
            "password": "password",
            "username": "shashank"
        }
    },
    "ackDeadlineSeconds": 15
}
```


internal/subscription/core.go
```diff
@@ -110,6 +112,44 @@ func (c *Core) CreateSubscription(ctx context.Context, m *Model) error {
...
+		logger.Ctx(ctx).Errorw("subscription not found", "name", uModel.Name)
+		return nil, merror.New(merror.NotFound, "subscription not found")
+	}
+
+	m, err := c.Get(ctx, uModel.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	err = patch.StructPatch(ctx, uModel, m, paths)
```
sunnyagg, line 145: should every model call patch pkg ?
can we abstract it out a bit

something like m.Patch(ctx, uModel, paths) 

and baseModel calling pkg and handling it, this way in future if we change the pkg we need to make changes at only one place

razorao: Let's say I define the method `Patch` in the IModel interface. And this will be implemented by the `BaseModel`. The `BaseModel` would implement the method.
`func (b *BaseModel) Patch(...){}`
`BaseModel` would be embedded inside evry other model. This would promote the implementation into each and every model. However, when invoked, this method would still receive the instance of BaseModel and not of the actual model. So the only other way to do this would be to implement it at every model level.. which is what we don't want.

shivasishdas:
@razorao Add the newly created API endpoint with request/response in the description for easy reference.

shivasishdas:
@razorao Did a quick setup of this branch on my local. This was the flow I had in mind using the already available `proto-helper-functions` and `fmutils` package.

```
// step 1
updateSubscriptionRequest.Fieldmask.Normalize()

// step 2
if !updateSubscriptionRequest.Fieldmask.IsValid(updateSubscriptionRequest.Subscription) {
		// Return an error.
}

// step 3 Redact the request according to the provided field mask.
fmutils.Filter(updateSubscriptionRequest.GetSubscription(), updateSubscriptionRequest.Fieldmask.GetPaths())

// step 4 Now that the request is vetted we can merge it with the Subscription entity.
proto.Merge(existingSubscriptionAsProto, updateProfileRequest.GetSubscription())
NOTE: savedDBSubscription to Proto and vice-versa is the additional needed mapping step.
```

Do have a look at these and see if these reduce custom code addition and increase re-usability across models for the future.

internal/subscription/core.go
sunnyagg: > this would be to implement it at every model level.. which is what we don't want.
Dont see issues here as long as we are not exposing it via Service/Core layer thought i get the point that BaseModel wont work, we need to solve it via composition to avoid duplication later

pkg/patch/patch.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	"github.com/oleiade/reflections"
+	"github.com/razorpay/metro/pkg/logger"
+)
+
+//StructPatch - Patches the attributes of struct `fromObj` into struct `toObj`
```
sunnyagg, line 12: nit: space after `//`

service/web/subscriberserver.go
```diff
@@ -47,6 +47,24 @@ func (s subscriberserver) CreateSubscription(ctx context.Context, req *metrov1.S
...
+	m, paths, err := subscription.GetValidatedModelAndPathsForUpdate(ctx, req)
+	if err != nil {
+		return nil, merror.ToGRPCError(err)
+	}
+
+	m, err = s.subscriptionCore.UpdateSubscription(ctx, m, paths)
+	if err != nil {
+		return nil, merror.ToGRPCError(err)
+	}
+	return subscription.ModelToSubscriptionProtoV1(m), nil
```
sunnyagg, line 65: can we just return req.subcription as response ?

internal/subscription/core.go
razorao: Will resolve conversation?

service/web/subscriberserver.go
razorao: > @razorao Did a quick setup of this branch on my local. This was the flow I had in mind using the already available `proto-helper-functions` and `fmutils` package.
> 
> ```
> // step 1
> updateSubscriptionRequest.Fieldmask.Normalize()
> 
> // step 2
> if !updateSubscriptionRequest.Fieldmask.IsValid(updateSubscriptionRequest.Subscription) {
> 		// Return an error.
> }
> 
> // step 3 Redact the request according to the provided field mask.
> fmutils.Filter(updateSubscriptionRequest.GetSubscription(), updateSubscriptionRequest.Fieldmask.GetPaths())
> 
> // step 4 Now that the request is vetted we can merge it with the Subscription entity.
> proto.Merge(existingSubscriptionAsProto, updateProfileRequest.GetSubscription())
> NOTE: savedDBSubscription to Proto and vice-versa is the additional needed mapping step.
> ```
> 
> Do have a look at these and see if these reduce custom code addition and increase re-usability across models for the future.

@shivasishdas Tried your suggestion
The problem is when we are tying to patch with an emtpy value, Eg: Patch an int with 0, or patch a pointer with nil,
The Merge method does not  apply the patch.

shivasishdas: @razorao Few questions:

- Whats happens on patch `int with 0` or `a string with ""`, do they simply get ignored? Asking as these seem to be default value cases for primitives.
- Does update work fine for non-default primitive values?

razorao:
> @razorao Did a quick setup of this branch on my local. This was the flow I had in mind using the already available `proto-helper-functions` and `fmutils` package.
> 
> ```
> // step 1
> updateSubscriptionRequest.Fieldmask.Normalize()
> 
> // step 2
> if !updateSubscriptionRequest.Fieldmask.IsValid(updateSubscriptionRequest.Subscription) {
> 		// Return an error.
> }
> 
> // step 3 Redact the request according to the provided field mask.
> fmutils.Filter(updateSubscriptionRequest.GetSubscription(), updateSubscriptionRequest.Fieldmask.GetPaths())
> 
> // step 4 Now that the request is vetted we can merge it with the Subscription entity.
> proto.Merge(existingSubscriptionAsProto, updateProfileRequest.GetSubscription())
> NOTE: savedDBSubscription to Proto and vice-versa is the additional needed mapping step.
> ```
> 
> Do have a look at these and see if these reduce custom code addition and increase re-usability across models for the future.


@shivasishdas Tried your suggestion
The problem is when we are tying to patch with an emtpy value, Eg: Patch an int with 0, or patch a pointer with nil,
The Merge method does not apply the patch.

shivasishdas:
> > @razorao Did a quick setup of this branch on my local. This was the flow I had in mind using the already available `proto-helper-functions` and `fmutils` package.
> > ```
> > // step 1
> > updateSubscriptionRequest.Fieldmask.Normalize()
> > 
> > // step 2
> > if !updateSubscriptionRequest.Fieldmask.IsValid(updateSubscriptionRequest.Subscription) {
> > 		// Return an error.
> > }
> > 
> > // step 3 Redact the request according to the provided field mask.
> > fmutils.Filter(updateSubscriptionRequest.GetSubscription(), updateSubscriptionRequest.Fieldmask.GetPaths())
> > 
> > // step 4 Now that the request is vetted we can merge it with the Subscription entity.
> > proto.Merge(existingSubscriptionAsProto, updateProfileRequest.GetSubscription())
> > NOTE: savedDBSubscription to Proto and vice-versa is the additional needed mapping step.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do have a look at these and see if these reduce custom code addition and increase re-usability across models for the future.
> 
> @shivasishdas Tried your suggestion
> The problem is when we are tying to patch with an emtpy value, Eg: Patch an int with 0, or patch a pointer with nil,
> The Merge method does not apply the patch.

@razorao Few questions:

- Whats happens on patch `int with 0` or `a string with ""`, do they simply get ignored? Asking as these seem to be default value cases for primitives. These problems exist with any struct update in Go. 
Ex An int field in a request struct will always get defaulted to 0. Same for booleans and strings.
- Does update work fine for non-default primitive values?

service/web/subscriberserver.go
razorao: patching int with 0 does not work, string I did not check, But I assume it's the same.
Update works fine for non empty values.. patching int from 100 -> 50 works, 100 -> 0 breaks
push_config: {"endpoint": "endpoint1"} -> push_config: {"endpoint": "endpoint2"} works
push_config: {"endpoint": "endpoint1"} -> nil, nope!

razorao:
> > > @razorao Did a quick setup of this branch on my local. This was the flow I had in mind using the already available `proto-helper-functions` and `fmutils` package.
> > > ```
> > > // step 1
> > > updateSubscriptionRequest.Fieldmask.Normalize()
> > > 
> > > // step 2
> > > if !updateSubscriptionRequest.Fieldmask.IsValid(updateSubscriptionRequest.Subscription) {
> > > 		// Return an error.
> > > }
> > > 
> > > // step 3 Redact the request according to the provided field mask.
> > > fmutils.Filter(updateSubscriptionRequest.GetSubscription(), updateSubscriptionRequest.Fieldmask.GetPaths())
> > > 
> > > // step 4 Now that the request is vetted we can merge it with the Subscription entity.
> > > proto.Merge(existingSubscriptionAsProto, updateProfileRequest.GetSubscription())
> > > NOTE: savedDBSubscription to Proto and vice-versa is the additional needed mapping step.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Do have a look at these and see if these reduce custom code addition and increase re-usability across models for the future.
> > 
> > 
> > @shivasishdas Tried your suggestion
> > The problem is when we are tying to patch with an emtpy value, Eg: Patch an int with 0, or patch a pointer with nil,
> > The Merge method does not apply the patch.
> 
> @razorao Few questions:
> 
> * Whats happens on patch `int with 0` or `a string with ""`, do they simply get ignored? Asking as these seem to be default value cases for primitives. These problems exist with any struct update in Go.
>   Ex An int field in a request struct will always get defaulted to 0. Same for booleans and strings.
> * Does update work fine for non-default primitive values?

@shivasishdas patching int with 0 does not work, string I did not check, But I assume it's the same.
Update works fine for non empty values.. patching int from 100 -> 50 works, 100 -> 0 breaks
push_config: {"endpoint": "endpoint1"} -> push_config: {"endpoint": "endpoint2"} works
push_config: {"endpoint": "endpoint1"} -> nil, nope!

internal/subscription/core.go
razorao: @sunnyagg So the current implementation works for now?

shivasishdas:
@razorao Check this out https://developers.google.com/protocol-buffers/docs/proto3 under ` Default Values` section

`Note that for scalar message fields, once a message is parsed there's no way of telling whether a field was explicitly set to the default value (for example whether a boolean was set to false) or just not set at all: you should bear this in mind when defining your message types. For example, don't have a boolean that switches on some behaviour when set to false if you don't want that behaviour to also happen by default. Also note that if a scalar message field is set to its default, the value will not be serialized on the wire.`

So the merge is working as expected actually.

razorao:
> @razorao Check this out https://developers.google.com/protocol-buffers/docs/proto3 under ` Default Values` section
> 
> `Note that for scalar message fields, once a message is parsed there's no way of telling whether a field was explicitly set to the default value (for example whether a boolean was set to false) or just not set at all: you should bear this in mind when defining your message types. For example, don't have a boolean that switches on some behaviour when set to false if you don't want that behaviour to also happen by default. Also note that if a scalar message field is set to its default, the value will not be serialized on the wire.`
> 
> So the merge is working as expected actually.

Right, so we cannot patch back default values..

shivasishdas:
> > @razorao Check this out https://developers.google.com/protocol-buffers/docs/proto3 under ` Default Values` section
> > `Note that for scalar message fields, once a message is parsed there's no way of telling whether a field was explicitly set to the default value (for example whether a boolean was set to false) or just not set at all: you should bear this in mind when defining your message types. For example, don't have a boolean that switches on some behaviour when set to false if you don't want that behaviour to also happen by default. Also note that if a scalar message field is set to its default, the value will not be serialized on the wire.`
> > So the merge is working as expected actually.
> 
> Right, so we cannot patch back default values using this..

So the purpose of FieldMask is to help us identify the fields of concern. If an `int` field is mentioned in the `update_path` list and the value is sent as `0`, then we should be fine with right?

razorao:
> > > @razorao Check this out https://developers.google.com/protocol-buffers/docs/proto3 under ` Default Values` section
> > > `Note that for scalar message fields, once a message is parsed there's no way of telling whether a field was explicitly set to the default value (for example whether a boolean was set to false) or just not set at all: you should bear this in mind when defining your message types. For example, don't have a boolean that switches on some behaviour when set to false if you don't want that behaviour to also happen by default. Also note that if a scalar message field is set to its default, the value will not be serialized on the wire.`
> > > So the merge is working as expected actually.
> > 
> > 
> > Right, so we cannot patch back default values using this..
> 
> So the purpose of FieldMask is to help us identify the fields of concern. If an `int` field is mentioned in the `update_path` list and the value is sent as `0`, then we should be fine with right?

If the int field is currently 100 and is being set to 0, then 0 must be patched. i.e now the field in registry must be set to 0. The `merge` method isn't doing it.

shivasishdas:
Ah! Checked the API doc now. Says `Populated scalar fields in src are copied to dst`. Default values are skipped 😞 
https://pkg.go.dev/github.com/golang/protobuf/proto#Merge

Can we look for an alternative for this? Pretty sure this would be a solved problem.

> If the int field is currently 100 and is being set to 0, then 0 must be patched. i.e now the field in registry must be set to 0. The `merge` method isn't doing it.

razorao:
> Ah! Checked the API doc now. Says `Populated scalar fields in src are copied to dst`. Default values are skipped 😞
> https://pkg.go.dev/github.com/golang/protobuf/proto#Merge
> 
> Found an open issue for this as well : [golang/protobuf#1127](https://github.com/golang/protobuf/issues/1127)
> 
> Can we look for an alternative for this? Pretty sure this would be a solved problem.
> 
> > If the int field is currently 100 and is being set to 0, then 0 must be patched. i.e now the field in registry must be set to 0. The `merge` method isn't doing it.

I've had bad luck with that :(
Had the same issue when searched for libraries that patch structs

service/web/subscriberserver.go
razorao: From what I've seen, google pub sub returns the actual data rather than what the client provides.
I tried out a scenario, where I do not provide the `ackDeadlineSeconds` attribute in the request (because I don't intend to update that). In the response, I receive the `ackDeadlineSeconds` as 100 (which is it's real value)

shivasishdas:
> I've had bad luck with that :(
> [golang/protobuf#1127 (comment)](https://github.com/golang/protobuf/issues/1127#issuecomment-628119126)
> Had the same issue when searched for libraries that patch structs

Check this lib out: https://github.com/imdario/mergo

razorao:
> > I've had bad luck with that :(
> > [golang/protobuf#1127 (comment)](https://github.com/golang/protobuf/issues/1127#issuecomment-628119126)
> > Had the same issue when searched for libraries that patch structs
> 
> Check this lib out: https://github.com/imdario/mergo

Works fine when combined with fmutils

sunnyagg (approved):

Issue #114, opened 2021-05-06
Edit subscription/Topic doesn't refresh the push subscription settings


sunnyagg:
Update subscription allows editing subscription endpoint and other properties which controls the messages delivery.
currently subscription update doesn't refresh these for a running push subscription.

PR #205, opened 2021-07-25
Issue #114: Metrics for Issue #114

razorao:
--Modified nodebinding model to store the time of scheduling
            --Added a metric to measure the latency between time of scheduing
              and time a task is picked for execution

### DIFF
--- a/internal/tasks/subscriptiontask.go
+++ b/internal/tasks/subscriptiontask.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sync"
+	"time"
 
 	"golang.org/x/sync/errgroup"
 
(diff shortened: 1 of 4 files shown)


shivasishdas:
@razorao Need to create panels as well for the metric. Ask for help from DevOps in case of access issues.
Non-prod: https://grafana.np.razorpay.in/d/FNNIEGQGz/metro-stage?orgId=1&from=now-24h&to=now
Prod: https://vajra.razorpay.com/dashboards/f/5oNBs5Rnk/prod-metro

internal/tasks/subscriptiontask.go
```diff
@@ -209,6 +210,9 @@ func (sm *SubscriptionTask) handleNodeBindingUpdates(ctx context.Context, newBin
 
 		if !found {
 			logger.Ctx(ctx).Infow("binding added", "key", newBinding.Key())
+			if !newBinding.ScheduledAt.IsZero() {
```
sunnyagg, line 213: This will only capture time to trigger a consul watch, we should take a diff with subscription updated_at to get the actual latency

razorao:
This should be handled using hypertrce

PR #206, opened 2021-07-26
Retry with delay buckets

shivasishdas:
**Feature:** Implements retry with pre-defined interval buckets

`Buckets used:` Delay5sec, Delay30sec, Delay60sec, Delay150sec, Delay300sec, Delay600sec, Delay1800sec, Delay3600sec

`Subscription changes`
- On subscription creation, the creation of subscription-specific delay topics is also done.
- `Subscription` model now additionally holds the names of all the created delay topics.

`Subscriber changes`
- On `Subscriber` init, a new `Retrier` is also constructed. This `Retrier` accepts the `Subscription`, `Backoff`, `IntervalFinder`, and a `MessageHandler` as inputs. 
- The `Subscriber` submits messages to be retried to the `Retrier`.

**New components below**

`Retrier`
- Spawns all the delay-consumers for a given list of delay topics.
- Given a broker message, finds the next closest delay topic to be used after exponential backoff and pushes the message onto the identified topic.

`Backoff`: Defines the backoff policy to be used to calculate the next interval.

`IntervalFinder`: Defines how to find an interval from a given slice of intervals.

`DelayConsumer`
- Each `DelayConsumer` is continuously polling off its assigned `DelayTopic`. 
- If a message is received which cannot be processed immediately, the `DelayConsumer` is paused. It gets resumed when the time elapses.
- Checks for message retry threshold and pushed to the dead-letter topic in case it is breached.

`MessageHandler`
- Defines the action to be taken once a message is processed by a `DelayConsumer`
- For now, only `PushToPrimaryRetryTopic` is supported.
- In future, more custom handlers can be onboarded by just complying with the `MessageHandler` interface.


internal/httpclient/httpclient.go
```diff
@@ -0,0 +1,37 @@
+package httpclient
```
sunnyagg, line 1: this can be part of pkg

internal/subscription/delay.go
```diff
@@ -0,0 +1,201 @@
...
+	upperBoundMaximumBackoffInSeconds uint = 600
+
+	defaultMaxDeliveryAttempts    int32 = 5
+	lowerBoundMaxDeliveryAttempts int32 = 5
+	upperBoundMaxDeliveryAttempts int32 = 100
+)
+
+var (
+	// ErrInvalidMinBackoff ...
+	ErrInvalidMinBackoff = errors.New("min backoff should be between 0 and 600 seconds")
```
razorao, line 30: Instead of hardcoding "0" "600" in the error string, use the constants `lowerBoundMinimumBackoffInSeconds` `upperBoundMinimumBackoffInSeconds` in the error string? This way, in case the ranges are changed in future, error string need not be explicitly changed.

internal/httpclient/httpclient.go
shivasishdas: moved

internal/subscription/delay.go
shivasishdas: Done.

razorao (approved):

internal/stream/stream.go
```diff
@@ -6,18 +6,17 @@ import (
...
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/opentracing/opentracing-go"
-	"golang.org/x/sync/errgroup"
-
 	"github.com/razorpay/metro/internal/subscriber"
 	"github.com/razorpay/metro/internal/subscription"
+	"github.com/razorpay/metro/pkg/httpclient"
 	"github.com/razorpay/metro/pkg/logger"
 	metrov1 "github.com/razorpay/metro/rpc/proto/v1"
+	"golang.org/x/sync/errgroup"
```
sunnyagg, line 19: nit: move this with the above list of imports

shivasishdas: done

internal/subscription/delay.go
```diff
@@ -0,0 +1,199 @@
...
+			"groupInstanceID": dc.GroupInstanceID,
+		},
+	}
+
+	fields = append(fields, kv...)
+	return fields
+}
+
+// DelayConfig ...
+type DelayConfig struct {
```
sunnyagg, line 91: do we need to duplicate this, we already have retry config and deadletter config in subscritpion.

internal/subscription/delay.go
```diff
@@ -0,0 +1,199 @@
...
+	MinDelay = Delay5sec
+	// MaxDelay ...
+	MaxDelay = Delay3600sec
+)
+
+// Intervals during subscription creation, query from the allowed intervals list, and create all the needed topics for retry.
+var Intervals = []Interval{Delay5sec, Delay30sec, Delay60sec, Delay150sec, Delay300sec, Delay600sec, Delay1800sec, Delay3600sec}
+
+// DelayConsumerConfig ...
+type DelayConsumerConfig struct {
```
sunnyagg, line 70: this seems bit off w.r.t subscription model namespace

internal/subscription/delay.go
```diff
@@ -0,0 +1,199 @@
...
+	// ErrInvalidMaxBackoff ...
+	ErrInvalidMaxBackoff = errors.New(fmt.Sprintf("max backoff should be between %v and %v seconds", lowerBoundMaximumBackoffInSeconds, upperBoundMaximumBackoffInSeconds))
+	// ErrInvalidMaxDeliveryAttempt ...
+	ErrInvalidMaxDeliveryAttempt = errors.New(fmt.Sprintf("max delivery attempt should be between %v and %v", lowerBoundMaxDeliveryAttempts, upperBoundMaxDeliveryAttempts))
+)
+
+// Interval is internal delay type per allowed interval
+type Interval uint
+
+var (
```
sunnyagg, line 40: feeling we should keep these in retry module

internal/subscription/delay.go
```diff
@@ -0,0 +1,199 @@
...
+// NewDelayConfig validates a subscription model and initializes the needed config values to be used for delay consumers
+func NewDelayConfig(model *Model) (*DelayConfig, error) {
+
+	minB := defaultMinimumBackoffInSeconds
+	maxB := defaultMaximumBackoffInSeconds
+	maxAtt := defaultMaxDeliveryAttempts
+
+	if model.RetryPolicy != nil {
+		minB = model.RetryPolicy.MinimumBackoff
+		if minB < lowerBoundMinimumBackoffInSeconds || minB > upperBoundMinimumBackoffInSeconds {
```
sunnyagg, line 111: this validation should be done with create/update subscription validations

internal/subscription/core.go
```diff
@@ -103,9 +101,37 @@ func (c *Core) CreateSubscription(ctx context.Context, m *Model) error {
 		})
 
 		if err != nil {
-			logger.Ctx(ctx).Errorw("failed to create deadletter topic for subscription", "name", m.GetDeadLetterTopic(), "error", err.Error())
+			logger.Ctx(ctx).Errorw("failed to create dead letter topic for subscription", "name", m.GetDeadLetterTopic(), "error", err.Error())
 			return err
 		}
+
+		// this creates the needed retry/delay topic configs
+		m.DelayConfig, err = NewDelayConfig(m)
```
sunnyagg, line 109: do we need a separate delay config in subscription model, we have retry/dlq configs already, most of the delayconfig is just copy of same

internal/subscription/delay.go
```diff
@@ -0,0 +1,199 @@
...
+			maxAtt = defaultMaxDeliveryAttempts
+		} else if maxAtt < lowerBoundMaxDeliveryAttempts || maxAtt > upperBoundMaxDeliveryAttempts {
+			return nil, ErrInvalidMaxDeliveryAttempt
+		}
+	}
+
+	delayTopics := make([]string, 0)
+	delayIntervalToTopicNameMap := make(map[Interval]DelayConsumerConfig)
+	groupID := uuid.New().String()
+	for _, interval := range Intervals {
```
sunnyagg, line 138: do we need to save names in consul here, can we generate then on runtime ?

internal/subscription/delay.go
```diff
@@ -0,0 +1,199 @@
...
+		delayTopic := topic.GetTopicName(model.ExtractedSubscriptionProjectID, fmt.Sprintf(delayTopicNameFormat, model.ExtractedSubscriptionName, interval))
+		delayTopics = append(delayTopics, delayTopic)
+		delayIntervalToTopicNameMap[interval] = DelayConsumerConfig{
```
sunnyagg, line 141: DelayConsumer logic should be handled by subscriber

internal/subscriber/core.go
```diff
@@ -39,36 +39,28 @@ func (c *Core) NewSubscriber(ctx context.Context,
...
-
-	dlqProducer, err := c.bs.GetProducer(ctx, messagebroker.ProducerClientOptions{Topic: dlqTopic, TimeoutMs: 500})
-	if err != nil {
-		logger.Ctx(ctx).Errorw("subscriber: failed to create dlq producer", "error", err.Error())
-		return nil, err
+	subsCtx, cancelFunc := context.WithCancel(ctx)
+	// using the subscriber ctx for retrier as well. This way when the ctx() for subscribers is done,
+	// all the delay-consumers spawned within retrier would also get marked as done.
+	var retrier IRetrier
+	if subscription.DelayConfig != nil {
```
sunnyagg, line 54: All subscription will have retry configuration, when do we not have this ?

internal/subscriber/core.go
```diff
@@ -39,36 +39,28 @@ func (c *Core) NewSubscriber(ctx context.Context,
...
+		retrier, err = NewRetrier(subsCtx, subscription.DelayConfig, c.bs, NewPushToPrimaryRetryTopicHandler(c.bs))
```
sunnyagg, line 55: do we need a retry per subscriber, can't we just have a retry module as singleton which manages all subscriptions and can retry a message based on message and subscription attributes

internal/subscription/core.go
```diff
@@ -103,9 +101,37 @@ func (c *Core) CreateSubscription(ctx context.Context, m *Model) error {
...
+		if err != nil {
+			logger.Ctx(ctx).Errorw("failed to init delay config", "error", err.Error())
+			return err
+		}
+
+		// create all the needed delay topics
+		for _, delayTopic := range m.DelayConfig.DelayTopics {
```
razorao, line 116: The part where we create all the delay topics can be inside a different function. This way, it can be reused tomorrow in `editSubscription` when retries are allowed to be edited.

razorao:
Created Issue #234 for edit of subscription retry

internal/subscriber/subscriber.go
```diff
@@ -87,78 +76,48 @@ func (s *Subscriber) GetID() string {
...
-	// remove message from the primary topic
-	_, err := s.consumer.CommitByPartitionAndOffset(ctx, messagebroker.CommitOnTopicRequest{
-		Topic:     retryMsg.Topic,
-		Partition: retryMsg.Partition,
-		// add 1 to current offset
-		// https://docs.confluent.io/5.5.0/clients/confluent-kafka-go/index.html#pkg-overview
-		Offset: retryMsg.Offset + 1,
-	})
-
+	// prepare message headers to be used by retrier
```
sunnyagg, line 91: can you add comments on what purpose each header is used for. looking at the current list not able to understand why we need so much data in headers, can just keeping subscription name do the trick ?

internal/subscriber/subscriber.go
```diff
@@ -87,78 +76,48 @@ func (s *Subscriber) GetID() string {
...
-
-	// then push message to the retry topic
-	_, err = s.retryProducer.SendMessage(ctx, messagebroker.SendMessageToTopicRequest{
-		Topic:      s.retryTopic,
-		Message:    retryMsg.Data,
-		TimeoutMs:  50,
-		MessageID:  retryMsg.MessageID,
-		RetryCount: retryMsg.incrementAndGetRetryCount(),
+	// commit on the primary topic after message has been submitted for retry
+	_, err = s.consumer.CommitByPartitionAndOffset(ctx, messagebroker.CommitOnTopicRequest{
```
sunnyagg, line 109: if you directly commit this, how does your in memory map updated, shouldn't commit happen as part of acknowledge ?

internal/subscriber/subscriber.go
```diff
@@ -410,36 +368,33 @@ func (s *Subscriber) Run(ctx context.Context) {
 		select {
 		case req := <-s.requestChan:
 			// on channel closure, we can get nil data
-			if req == nil {
-				continue
-			}
-			if ctx.Err() != nil {
+			if req == nil || ctx.Err() != nil {
```
sunnyagg, line 371: why is ctx.Err check needed, that is handlled by ctx.Done ?

sunnyagg (changes requested):
1. Almost 0 coverage of new code
2. Need to discuss interfaces, current structure seems complicating things
3. Migration of existing data is not handled, how do existing subscription moved to new retry behaviour, what is needed to support that

internal/subscriber/subscriber.go
```diff
@@ -208,9 +167,8 @@ func (s *Subscriber) acknowledge(req *AckMessage) {
 
 		logger.Ctx(ctx).Infow("subscriber: msg hit deadline", "logFields", logFields)
 
-		// push to retry queue
-		s.retry(ctx, NewRetryMessage(msg.Topic, msg.Partition, msg.Offset, msg.Data, msgID, msg.RetryCount))
-
+		// push for retry
+		s.retry(ctx, msg)
 		s.removeMessageFromMemory(ctx, stats, req.MessageID)
```
sunnyagg, line 172: removing will mean that message is acknowledged as well, which is incorrect, its still being retried, in memory removal shouldn't happen until we move to DLQ

internal/subscriber/subscriber.go
```diff
@@ -87,78 +76,48 @@ func (s *Subscriber) GetID() string {
 
 // GetSubscription ...
 func (s *Subscriber) GetSubscription() string {
```
razorao, line 78: Since `subscription` in subscriber is a subscription model and we are returning the subscription name here, rename this to `GetSubscriptionName` ?

internal/subscriber/core.go
shivasishdas: old subscriptions will not have this. Will need to do a migration for this. Capturing the same in the Tech Spec.

internal/subscription/delay.go
shivasishdas: So DelayConfig is a list of all delay topics created for the subscription based on the intervals that were available at that time. Need to keep this as a separate config because if we modify the interval buckets in future, this subscription will remain unaffected and the change will apply only to the new ones.

internal/subscription/delay.go
shivasishdas: suggestions?

internal/subscription/delay.go
shivasishdas: This is getting called on subscriptions create. Are you suggesting moving this validation inside subscription/validations?

internal/subscription/core.go
shivasishdas: This was created to keep track of the delay-topics created for a subscripition.

internal/subscription/delay.go
shivasishdas: We are creating delay-topics here. Seems simpler to save the topic names as well.
And I have explained the reason to store the delay-topic names for a subscription (as delay intervals can be changes in future)

internal/subscription/delay.go
shivasishdas: Agreed. There is no logic here. Its just converting a slice of topic names to map for quick lookups.

internal/subscriber/core.go
shivasishdas: Backward compatibitliy. Old subscriptions will not have this. Need to run a migration for them. Will cover in spec.

internal/subscriber/core.go
shivasishdas: We had decided during our Spec discussion that each subscriber/subscription should get its own retry module.
- This way the retry module handles only the subscription-specific traffic and one topic experiencing high retry messages will not affect the throughput of another subscription. 
- Additionally makes it easier to break for ingestion topics in the future.

internal/subscription/core.go
shivasishdas: agreed. done.

internal/subscriber/subscriber.go
shivasishdas: For every `s.retry()` call, we call a subsequent `s.removeMessageFromMemory(ctx, stats, req.MessageID)` from the calling function. This takes care of cleaning up the in-memory ds.

internal/subscriber/subscriber.go
shivasishdas: added.

internal/subscriber/subscriber.go
shivasishdas: Was seeing a case where the loop enters the case block and then the ctx() gets canceled.

internal/subscriber/retrier.go
```diff
@@ -0,0 +1,159 @@
...
+// NewRetrier inits a new retrier which internally takes care of spawning the needed delay-consumers.
+func NewRetrier(ctx context.Context, dc *subscription.DelayConfig, bs brokerstore.IBrokerStore, handler RetryMessageHandler) (IRetrier, error) {
+	delayConsumers := sync.Map{}
+
+	for interval, config := range dc.GetDelayTopicsMap() {
+		dc, err := NewDelayConsumer(ctx, config, bs, handler)
+		if err != nil {
+			return nil, err
+		}
+		go dc.Run(ctx)                     // run the delay consumer
```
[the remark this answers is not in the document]
shivasishdas: So a retrier will currently run when a subscriber runs for a subscription.

> Can we explicitly call a start method 

Can you help me with more details on why this approach?

internal/subscription/delay.go
```diff
@@ -0,0 +1,199 @@
...
+
+var (
+	// MinDelay ...
+	MinDelay = Delay5sec
+	// MaxDelay ...
+	MaxDelay = Delay3600sec
+)
+
+// Intervals during subscription creation, query from the allowed intervals list, and create all the needed topics for retry.
+var Intervals = []Interval{Delay5sec, Delay30sec, Delay60sec, Delay150sec, Delay300sec, Delay600sec, Delay1800sec, Delay3600sec}
```
shivasishdas, line 67: EXPLORE: Kafka auto-create. This way we do not need to pre-create topics and manage them as a config.

internal/subscriber/core.go
```diff
@@ -39,36 +40,28 @@ func (c *Core) NewSubscriber(ctx context.Context,
...
-
-	dlqProducer, err := c.bs.GetProducer(ctx, messagebroker.ProducerClientOptions{Topic: dlqTopic, TimeoutMs: 500})
-	if err != nil {
-		logger.Ctx(ctx).Errorw("subscriber: failed to create dlq producer", "error", err.Error())
-		return nil, err
+	subsCtx, cancelFunc := context.WithCancel(ctx)
+	// using the subscriber ctx for retrier as well. This way when the ctx() for subscribers is done,
+	// all the delay-consumers spawned within retrier would also get marked as done.
+	var retrier retry.IRetrier
+	if subscription.RetryPolicy != nil && subscription.DeadLetterPolicy != nil {
```
sunnyagg, line 55: these should never be null, can we handle it gracefully for older subscriptions.
i.e  subscription.GetRetryPolicy() returns default policy if its not set and same for deadletterpolicy

once migrated we can remove the nulll checks or let them be there for safety

internal/subscriber/retry/delayconsumer.go
```diff
@@ -0,0 +1,186 @@
...
+		Topic:     msg.DeadLetterTopic,
+		TimeoutMs: defaultBrokerOperationsTimeoutMs,
+	})
+	if err != nil {
+		return err
+	}
+
+	_, err = dlProducer.SendMessage(dc.ctx, messagebroker.SendMessageToTopicRequest{
+		Topic:         msg.DeadLetterTopic,
+		Message:       msg.Data,
```
sunnyagg, line 119: we should add some attributes to the message, pubsub adds few attributes related to source topic details in the deadletter topic

internal/subscriber/retry/delayconsumer.go
```diff
@@ -0,0 +1,186 @@
...
+		} else {
+			// submit to handler
+			err := dc.handler.Do(dc.ctx, msg)
+			if err != nil {
+				logger.Ctx(dc.ctx).Errorw("delay-consumer: error in msg handler", dc.LogFields("error", err.Error())...)
+				return
+			}
+		}
+
+		// commit the message
```
sunnyagg, line 154: commit should be taken care by ack, if we commit here we wont be able to retry on failure

internal/subscriber/retry/retrier.go
```diff
@@ -0,0 +1,172 @@
...
+	})
+	wg.Wait()
+}
+
+// Handle takes care of processing a given retry message.
+func (r *Retrier) Handle(ctx context.Context, msg messagebroker.ReceivedMessage) error {
+
+	logger.Ctx(ctx).Infow("retrier: received msg for retry", msg.LogFields()...)
+
+	availableDelayIntervals := subscription.Intervals
```
sunnyagg, line 77: move intervals constants under retry module ?

internal/subscriber/retry/retrier.go
```diff
@@ -0,0 +1,172 @@
...
+
+	nextDelayInterval := calculateNextUsingExponentialBackoff(float64(msg.InitialDelayInterval), float64(msg.CurrentDelayInterval), float64(msg.CurrentRetryCount))
```
sunnyagg, line 79: `calculateNextUsingExponentialBackoff` this method needs to be part of interface which can have different implementations

internal/subscription/delay.go
```diff
@@ -0,0 +1,121 @@
+package subscription
```
sunnyagg, line 1: most of the constants belong to retry module we should move them there, for validations we should use validation.go file

internal/subscriber/core.go
shivasishdas: Doing a fallback only on default Retry and PL policies will not suffice as we need to have the delay topics created as well for retries to work. Ideally, the migration step should take care of old subscriptions and all new subscriptions after this PR should have non-nil values.

internal/subscriber/retry/delayconsumer.go
shivasishdas: Agreed! Wasn't sure of the fields to be published in the outgoing message. Can you share a sample gPubSub message with the needed fields for reference?

internal/subscriber/retry/retrier.go
shivasishdas: I think this will cause a cyclic dependency.

internal/subscription/delay.go
shivasishdas: Will cause cyclic dependency.

internal/subscriber/retry/delayconsumer.go
shivasishdas: I see that [delivery_attempts](https://github.com/razorpay/metro/pull/245) has already been added by @razorao. Anymore needed?

internal/subscriber/retry/delayconsumer.go
shivasishdas: This is a commit on the delay topic. `ack()` only has access to primary and primary-retry topics.

internal/subscriber/retry/retrier.go
shivasishdas: made changes

sunnyagg (approved):

Issue #183, opened 2021-06-23
Filter out status check endpoints from tracing


shivasishdas:
anyways will make this change in a different PR. Need to figure out the middleware interceptor.

_Originally posted by @shivasishdas in https://github.com/razorpay/metro/pull/180#discussion_r656416327_

sunnyagg:
@razorao close this ?

PR #207, opened 2021-07-27
Issue #183:  Prevent health check apis from opentracing

razorao:
Added a filter to grpcopentracing's Interceptor to filter out the health check APIs

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -134,7 +134,13 @@ func newGrpcServer(r registerGrpcHandlers, interceptors ...grpc.UnaryServerInter
 		// Add tags to logger context.
 		grpcinterceptor.UnaryServerLoggerInterceptor(),
 		// Todo: Confirm tracing is working as expected. Jaegar integration?
-		grpcopentracing.UnaryServerInterceptor(),
+		grpcopentracing.UnaryServerInterceptor(grpcopentracing.WithFilterFunc(func(ctx context.Context, fullMethodName string) bool {
+			if fullMethodName == "/google.pubsub.v1.StatusCheckAPI/LivenessCheck" ||
+				fullMethodName == "/google.pubsub.v1.StatusCheckAPI/ReadinessCheck" {
+				return false
+			}
+			return true
+		})),
 		// Instrument prometheus metrics for all methods. This will have a counter & histogram of latency.
 		grpcprometheus.UnaryServerInterceptor,
 	}
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -21,6 +21,18 @@ import (
 	"github.com/razorpay/metro/pkg/logger"
 )
 
+// Map containing the gRPC methods that should not be traced
+// Keys are the gRPC method fullnames
+// If method is present in map, tracing will be disabled
+var tracingExcludedMethods map[string]bool
+
+func init() {
+	tracingExcludedMethods = map[string]bool{
+		"/google.pubsub.v1.StatusCheckAPI/LivenessCheck":  true,
+		"/google.pubsub.v1.StatusCheckAPI/ReadinessCheck": true,
+	}
+}
+
 type registerGrpcHandlers func(server *grpc.Server) error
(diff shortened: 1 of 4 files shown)


internal/server/server.go
```diff
@@ -134,7 +134,13 @@ func newGrpcServer(r registerGrpcHandlers, interceptors ...grpc.UnaryServerInter
 		// Add tags to logger context.
 		grpcinterceptor.UnaryServerLoggerInterceptor(),
 		// Todo: Confirm tracing is working as expected. Jaegar integration?
-		grpcopentracing.UnaryServerInterceptor(),
+		grpcopentracing.UnaryServerInterceptor(grpcopentracing.WithFilterFunc(func(ctx context.Context, fullMethodName string) bool {
+			if fullMethodName == "/google.pubsub.v1.StatusCheckAPI/LivenessCheck" ||
```
shivasishdas, line 138: better to maintain a running list of URLs as a constant somewhere. This list may/can expand in the future.

internal/server/server.go
```diff
@@ -134,7 +134,13 @@ func newGrpcServer(r registerGrpcHandlers, interceptors ...grpc.UnaryServerInter
 		// Add tags to logger context.
 		grpcinterceptor.UnaryServerLoggerInterceptor(),
 		// Todo: Confirm tracing is working as expected. Jaegar integration?
-		grpcopentracing.UnaryServerInterceptor(),
+		grpcopentracing.UnaryServerInterceptor(grpcopentracing.WithFilterFunc(func(ctx context.Context, fullMethodName string) bool {
```
shivasishdas, line 137: Explore possible UT?

shivasishdas (changes requested):

internal/server/server.go
```diff
@@ -196,3 +196,19 @@ func newInternalServer() (*http.Server, error) {
 
 	return &server, nil
 }
+
+// traceMethod - determines if opentracing should be enabled for the grpc method
+func traceMethod(ctx context.Context, fullMethodName string) bool {
+	donotTraceMethods := []string{ // Should be a very small list
```
shivasishdas, line 202: nit: I understand the list is small for now, but still better to keep this as a map here for a constant time lookup.

internal/server/server.go
```diff
@@ -196,3 +196,19 @@ func newInternalServer() (*http.Server, error) {
 
 	return &server, nil
 }
+
+// traceMethod - determines if opentracing should be enabled for the grpc method
+func traceMethod(ctx context.Context, fullMethodName string) bool {
```
shivasishdas, line 201: a better name like `shouldEnableTrace`?

internal/server/server.go
shivasishdas: that being said, can this being a `var map[string]bool` init'ed on boot instead of a separate function?

shivasishdas (changes requested):

sunnyagg (approved):

internal/server/server.go
```diff
@@ -196,3 +208,9 @@ func newInternalServer() (*http.Server, error) {
 
 	return &server, nil
 }
+
+// shouldEnableTrace - determines if opentracing should be enabled for the grpc method
+func shouldEnableTrace(ctx context.Context, fullMethodName string) bool {
```
shivasishdas, line 213: nit: Actually, the function determines which path to exclude from the trace. So rename to shouldDisableTrace or shouldSkipTrace

shivasishdas (changes requested):

internal/server/server.go
razorao: Check the implementation of `UnaryServerInterceptor`
The filter method needs to return `false` to skip tracing
Hence.. Should enable trace? `false` = no

shivasishdas (approved):

razorao:
@sunnyagg @shivasishdas  Have added the environment as part of the service name in tracer configuration.

shivasishdas (approved):

Issue #228, opened 2021-08-17
Fix Authorization in GRPC


razorao:
Currently, project id is fetched from the URI and is validated against users's project id to perform authorization. However, in case of GRPC, the URI does not exist leading to breakage of authorization.

PR #229, opened 2021-08-17
Issue #228: Fix authorization in gRPC Calls

razorao:
- Authorization in gRPC calls was broken because _project-id_ was being extracted from the URL and compared against the user's project-id. 
- Implemented a custom `UnaryServerAuthInterceptor` and `serviceAuthFuncOverride` _interface_ which declares the `AuthFuncOverride` _function_. If the server (read controller) implements the interface function, the override function is called with the request body and the full function name. 
- This implementation is similar to `grpc_auth.UnaryServerInterceptor`, except that the override function gets the request body.
- Added the file _metro/service/web/server.go_ which defines methods to extract project-id from the request body. 
- The project-id is extracted from the _resource-name_ such as _topic-name_, _subscription-name_ using _regex capturing-group_.


internal/interceptors/auth.go
```diff
@@ -33,9 +50,9 @@ func getUserPasswordProjectID(ctx context.Context) (user string, password []byte
 
 	// `uri` gets set inside `runtime.WithMetadata()`. check `server.go` for implementation
 	// this is ideally a single-valued slice
-	if val := md.Get("uri"); val != nil && len(val) > 0 {
-		projectID = extractProjectIDFromURI(val[0])
-	}
+	// if val := md.Get("uri"); val != nil && len(val) > 0 {
```
sunnyagg, line 53: remove commented code

service/web/server.go
```diff
@@ -0,0 +1,66 @@
...
+func getProjectIDFromRequest(ctx context.Context, req interface{}) (string, error) {
+	resourceName, err := getResourceNameFromRequest(ctx, req)
+	if err != nil {
+		return "", err
+	}
+	return getProjectIDFromResourceName(ctx, resourceName)
+}
+
+func getResourceNameFromRequest(ctx context.Context, req interface{}) (string, error) {
+	switch t := req.(type) {
```
sunnyagg, line 29: instead of request type, should we pass grpc method name and based on method name switch case, use appropriate type to desrialize

service/web/server.go
```diff
@@ -0,0 +1,66 @@
...
+		return req.(*metrov1.UpdateSubscriptionRequest).Subscription.Name, nil
+	case *metrov1.AcknowledgeRequest:
+		return req.(*metrov1.AcknowledgeRequest).Subscription, nil
+	case *metrov1.PullRequest:
+		return req.(*metrov1.PullRequest).Subscription, nil
+	case *metrov1.DeleteSubscriptionRequest:
+		return req.(*metrov1.DeleteSubscriptionRequest).Subscription, nil
+	case *metrov1.ModifyAckDeadlineRequest:
+		return req.(*metrov1.ModifyAckDeadlineRequest).Subscription, nil
+	default:
```
sunnyagg, line 48: what about streaming pull

sunnyagg (approved):

service/web/server.go
razorao: We can do that. But a given proto request can be used by multiple methods. Hence using type switch will give us less switch cases

service/web/server.go
```diff
@@ -0,0 +1,66 @@
...
+	case *metrov1.PublishRequest:
+		return req.(*metrov1.PublishRequest).Topic, nil
+	case *metrov1.Topic:
+		return req.(*metrov1.Topic).Name, nil
+	case *metrov1.DeleteTopicRequest:
+		return req.(*metrov1.DeleteTopicRequest).Topic, nil
+	case *metrov1.Subscription:
+		return req.(*metrov1.Subscription).Name, nil
+	case *metrov1.UpdateSubscriptionRequest:
+		return req.(*metrov1.UpdateSubscriptionRequest).Subscription.Name, nil
```
shivasishdas, line 39: nested access can result in nil panic if `Subscription` is sent as null in the `json`. Better to have a nil check-here

service/web/server.go
```diff
@@ -0,0 +1,66 @@
...
+		return req.(*metrov1.AcknowledgeRequest).Subscription, nil
+	case *metrov1.PullRequest:
+		return req.(*metrov1.PullRequest).Subscription, nil
+	case *metrov1.DeleteSubscriptionRequest:
+		return req.(*metrov1.DeleteSubscriptionRequest).Subscription, nil
+	case *metrov1.ModifyAckDeadlineRequest:
+		return req.(*metrov1.ModifyAckDeadlineRequest).Subscription, nil
+	default:
+		logger.Ctx(ctx).Infof("unknown request type: %v", t)
+		err := merror.New(merror.Unimplemented, "unknown resource type")
```
shivasishdas, line 50: declare errors as `var` on the top of the file. Easier to compare and validate in UTs. check other custom errors as well in the PR.

service/web/server_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"context"
+	"testing"
+
+	metrov1 "github.com/razorpay/metro/rpc/proto/v1"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetProjectIDFromRequest_PublishRequest(t *testing.T) {
```
shivasishdas, line 11: can refactor the tests as table-driven and re-use the assertions. Example below.
Refer https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

```
type test struct {
	input    interface{}
	expected string
	err      error
}

	tests := []test{
		{
			&metrov1.PublishRequest{
				Topic: "projects/project-p001/topics/topic-t001",
			}, "project-p001", nil,
		},
		{
			&metrov1.UpdateSubscriptionRequest{
				Subscription: &metrov1.Subscription{
					Name: "projects/project-p001/subscriptions/subscription-s001",
				},
			}, "project-p001", nil,
		},
	}

	for _, ts := range tests {
		pid, err := getProjectIDFromRequest(context.Background(), ts.input)
		assert.Equal(t, ts.expected, pid)
		assert.Equal(t, ts.err, err)
	}
```

service/web/subscriberserver.go
```diff
@@ -229,6 +229,11 @@ func (s subscriberserver) ModifyAckDeadline(ctx context.Context, req *metrov1.Mo
 	return new(emptypb.Empty), nil
 }
 
-func (s subscriberserver) AuthFuncOverride(ctx context.Context, _ string) (context.Context, error) {
-	return interceptors.AppAuth(ctx, s.credentialCore)
+//AuthFuncOverride - Override function called by the auth interceptor
+func (s subscriberserver) AuthFuncOverride(ctx context.Context, _ string, req interface{}) (context.Context, error) {
+	projectID, err := getProjectIDFromRequest(ctx, req)
```
shivasishdas, line 234: we are duplicating these lines in each server for now. Let's try and make this common. Maybe wrap `interceptors.AppAuth(ctx, s.credentialCore, projectID)` in another func call and call that from each server impl.

internal/interceptors/auth_test.go
```diff
@@ -194,6 +186,6 @@ func Test_AppAuth_Failure5(t *testing.T) {
 	mockCore := mocks1.NewMockICore(ctrl)
 	mockCore.EXPECT().Get(gomock.Any(), project, username).Return(dummyCreds, nil)
 
-	ctx, err := AppAuth(newCtx, mockCore)
+	ctx, err := AppAuth(newCtx, mockCore, "")
```
shivasishdas, line 189: This test case is checking for `wrong password sent in authorizationHeaderKey` being sent. Let's keep the intent of the `TC` clear and send a non-empty valid `projectID` here.
Check other places as well in the file. Add failures on wrong/empty `projectID` as separate tests and make sure those tests specifically fail only the `projectID` checks.

shivasishdas (changes requested):

service/web/server.go
razorao: Have created #231 for this. Will take it up in a different PR.

shivasishdas (approved):
LGTM

PR #230, opened 2021-08-17
commenting out rotate publisher

shivasishdas:
This was added as a work-around to handle publishers timing out. The issue was with an infra config which has been identified and fixed.

### DIFF
--- a/internal/publisher/core.go
+++ b/internal/publisher/core.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/confluentinc/confluent-kafka-go/kafka"
 	"github.com/golang/protobuf/proto"
 	"github.com/opentracing/opentracing-go"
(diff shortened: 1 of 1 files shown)


internal/publisher/core.go
```diff
@@ -61,10 +60,10 @@ func (p *Core) Publish(ctx context.Context, req *metrov1.PublishRequest) ([]stri
 			OrderingKey: msg.OrderingKey,
 		})
 		if err != nil {
-			if err.Error() == kafka.ErrMsgTimedOut.String() {
-				logger.Ctx(ctx).Infow("got error, rotating producer", "error", err.Error())
-				p.bs.RemoveProducer(ctx, producerOps)
-			}
+			//if err.Error() == kafka.ErrMsgTimedOut.String() {
```
sunnyagg, line 63: add comments on this with TODO

sunnyagg (approved):

PR #236, opened 2021-08-31
Set the content type in metro push

razorao:
--Metro push requests were being made without content-type header
	--Setting the content type of push request to application/json
	--Log the response body in case of non 2xx responses


internal/stream/stream.go
```diff
@@ -200,6 +201,14 @@ func (ps *PushStream) pushMessage(ctx context.Context, subModel *subscription.Mo
 	}
 
 	// discard response.Body after usage and ignore errors
+	if !success {
+		bodyBytes, err := ioutil.ReadAll(resp.Body)
```
shivasishdas, line 205: Let's discuss if we should log this or not. Can contain sensitive information.

shivasishdas (approved):

Issue #234, opened 2021-08-20
Allow editing subscription retry details


razorao:
Currently, only push endpoint and messageAckSecs are editable attributes

PR #240, opened 2021-09-06
subscriptions migrate api

shivasishdas:
Proto defs: https://github.com/razorpay/metro-proto/pull/24/files

### DIFF
--- a/internal/subscription/core.go
+++ b/internal/subscription/core.go
@@ -278,6 +278,10 @@ func (c *Core) Get(ctx context.Context, key string) (*Model, error) {
 }
 
 // Migrate takes care of migrating the old subscription model to new
+// As of now it specifically takes care of two things:
+// 1. Updating the missing retry and dead-letter policies with default values.
+// 2. Create the needed delay topics for a subscription retry to work.
+// In future, modify this function as needed to support additional migration use-cases.
 func (c *Core) Migrate(ctx context.Context, names []string) error {
 
 	subscriptionsToUpdate := make([]*Model, 0)
--- a/internal/subscription/core.go
+++ b/internal/subscription/core.go
@@ -301,24 +301,30 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
 
 	topicNames := make([]string, 0)
 	for _, model := range subscriptionsToUpdate {
+		needsUpdate := false
+
 		// update retry policy if not set
 		if model.RetryPolicy == nil {
 			model.setDefaultRetryPolicy()
+			needsUpdate = true
 		}
 
 		// update dead letter policy if not set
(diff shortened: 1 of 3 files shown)


internal/subscription/core.go
```diff
@@ -22,6 +22,7 @@ type ICore interface {
 	ListKeys(ctx context.Context, prefix string) ([]string, error)
 	List(ctx context.Context, prefix string) ([]*Model, error)
 	Get(ctx context.Context, key string) (*Model, error)
+	Migrate(ctx context.Context, names []string) error
```
sunnyagg, line 25: what about updates happening later, lets say we add more retry durations ? does update api handles updating the topics ?

sunnyagg (approved):

internal/subscription/core.go
shivasishdas: That should go via the normal update subscription route. Already tracking issue here https://github.com/razorpay/metro/issues/234

razorao: Isn't the function name Migrate a bit generic? We might want to do another kind of migration in future to a different subscription model. In that case will we write another function or are we planning to change the current function's implementation?

internal/subscription/core.go
```diff
@@ -275,3 +276,75 @@ func (c *Core) Get(ctx context.Context, key string) (*Model, error) {
 	}
 	return model, nil
 }
+
+// Migrate takes care of migrating the old subscription model to new
+func (c *Core) Migrate(ctx context.Context, names []string) error {
+
+	subscriptionsToUpdate := make([]*Model, 0)
+	if names == nil || len(names) == 0 {
+		models, err := c.List(ctx, Prefix)
```
razorao, line 285: So if there is an empty list, all subscriptions are picked up. Currently we are assuming that there are very few subscriptions. But in case if the list is big, the API might take significant amount of time? Does this not pose any problem ?

internal/subscription/core.go
shivasishdas: Sure. I thought of this as throw-away code while writing this to solve the problem at hand. We should be able to modify it as needed in the future as more such use-cases creep in.

internal/subscription/core.go
shivasishdas: I had called it out during the Tech Spec discussion. Do refer `Assumptions` section for `Solution#1` under 'Migration of old subscriptions' in the Spec. This current implementation is because we know for sure the number of subscriptions is less in production.

shivasishdas: Have added a more detailed function description now. Do check and let me know if this suffices for the time being.

razorao (approved):

PR #242, opened 2021-09-07
Added validation and error handling for ackId

### LINKED ISSUE #227: Panic when parsing acknowledgement id
In the methid `ParseAckID`, we don't check if the ack id is valid. We split the ack id and directly access indices 0-6. If the ack id is invalid, this causes a panic and brings down the webserver.

rutul25:
- Added basic validation for ackId before attempting to parsing it to make sure expected number of components are there in the string
- Added error handling in case of any other run time failures to make sure code does not go into panic

### DIFF
--- a/internal/stream/stream.go
+++ b/internal/stream/stream.go
@@ -240,7 +240,11 @@ func (ps *PushStream) nack(ctx context.Context, message *metrov1.ReceivedMessage
 	logFields["ackId"] = message.AckId
 
 	logger.Ctx(ctx).Infow("worker: sending nack request to subscriber", "logFields", logFields)
-	ackReq := subscriber.ParseAckID(message.AckId)
+	ackReq, err := subscriber.ParseAckID(message.AckId)
+	if err != nil {
+		logger.Ctx(ctx).Errorf("worker: error in parsing ackId", "error", err.Error(), "logFields", logFields)
+		return
+	}
 	// deadline is set to 0 for nack
 	modackReq := subscriber.NewModAckMessage(ackReq, 0).WithContext(ctx)
 	// check for closed channel before sending request
@@ -271,7 +275,13 @@ func (ps *PushStream) ack(ctx context.Context, message *metrov1.ReceivedMessage)
 	logFields["ackId"] = message.AckId
 
 	logger.Ctx(ctx).Infow("worker: sending ack request to subscriber", "logFields", logFields)
-	ackReq := subscriber.ParseAckID(message.AckId).WithContext(ctx)
+	ackReq, err := subscriber.ParseAckID(message.AckId)
+	if err != nil {
+		logger.Ctx(ctx).Errorw("worker: error in parsing ackId", "error", err.Error(), "logFields", logFields)
+		return
+	}
+	ackReq = ackReq.WithContext(ctx)
+
 	// check for closed channel before sending request
 	if ps.subs.GetAckChannel() != nil {
 		ps.subs.GetAckChannel() <- ackReq
--- a/internal/subscriber/model.go
+++ b/internal/subscriber/model.go
@@ -10,10 +10,11 @@ import (
 	"strings"
 	"time"
 
-	"github.com/razorpay/metro/pkg/utils"
-
+	"github.com/pkg/errors"
+	"github.com/razorpay/metro/internal/merror"
 	"github.com/razorpay/metro/internal/subscriber/customheap"
 	"github.com/razorpay/metro/pkg/messagebroker"
+	"github.com/razorpay/metro/pkg/utils"
 )
 
 // PullRequest ...
@@ -78,11 +79,31 @@ type AckMessage struct {
 	AckID         string
 }
 
+var (
+	errIllegalPartitionValue = errors.New("partition value cannot be less than 0")
+	errIllegalOffsetValue    = errors.New("offset value cannot be less than 0")
+	errIllegalDeadlineValue  = errors.New("deadline value cannot be less than 0")
+	errInvalidAckID          = merror.Newf(merror.InvalidArgument, "AckID received is not in expected format")
+)
+
 const ackIDSeparator = "_"
 
 // NewAckMessage ...
-func NewAckMessage(subscriberID, topic string, partition, offset, deadline int32, messageID string) IAckMessage {
-	// TODO: add needed validations on all fields
+func NewAckMessage(subscriberID, topic string, partition, offset, deadline int32, messageID string) (IAckMessage, error) {
+
+	// Validating parameters for illegal values
+	if partition < 0 {
+		return nil, errIllegalPartitionValue
+	}
+
+	if offset < 0 {
+		return nil, errIllegalOffsetValue
+	}
+
+	if deadline < 0 {
+		return nil, errIllegalDeadlineValue
+	}
+
 	return &AckMessage{
 		SubscriberID: subscriberID,
 		Topic:        topic,
@@ -90,7 +111,7 @@ func NewAckMessage(subscriberID, topic string, partition, offset, deadline int32
 		MessageID:    messageID,
 		Deadline:     deadline,
 		Offset:       offset,
-	}
+	}, nil
 }
 
 // WithContext can be used to set the current context to the request
@@ -151,25 +172,63 @@ func (a *AckMessage) HasHitDeadline() bool {
 }
 
 // ParseAckID ...
-func ParseAckID(ackID string) *AckMessage {
+func ParseAckID(ackID string) (*AckMessage, error) {
 	// split the message and parse tokens
 	parts := strings.Split(ackID, ackIDSeparator)
+	decodedParts := decode(parts)
 
-	// TODO : add validations
-	partition, _ := strconv.ParseInt(utils.Decode(parts[3]), 10, 0)
-	offset, _ := strconv.ParseInt(utils.Decode(parts[4]), 10, 0)
-	deadline, _ := strconv.ParseInt(utils.Decode(parts[5]), 10, 0)
+	if !isValidAckID(decodedParts) {
+		return nil, errInvalidAckID
+	}
+
+	partition, err := strconv.ParseInt(decodedParts[3], 10, 0)
+	if err != nil {
+		err = errors.WithMessage(err, fmt.Sprintf("error in parsing partition value %v to int", decodedParts[3]))
+		return nil, err
+	}
+
+	offset, err := strconv.ParseInt(decodedParts[4], 10, 0)
+	if err != nil {
+		err = errors.WithMessage(err, fmt.Sprintf("error in parsing offset value %v to int", decodedParts[4]))
+		return nil, err
+	}
+
+	deadline, err := strconv.ParseInt(decodedParts[5], 10, 0)
+	if err != nil {
+		err = errors.WithMessage(err, fmt.Sprintf("error in parsing deadline value %v to int", decodedParts[5]))
+		return nil, err
+	}
 
 	return &AckMessage{
-		ServerAddress: utils.Decode(parts[0]),
-		SubscriberID:  utils.Decode(parts[1]),
-		Topic:         utils.Decode(parts[2]),
+		ServerAddress: decodedParts[0],
+		SubscriberID:  decodedParts[1],
+		Topic:         decodedParts[2],
 		Partition:     int32(partition),
 		Offset:        int32(offset),
 		Deadline:      int32(deadline),
-		MessageID:     utils.Decode(parts[6]),
+		MessageID:     decodedParts[6],
 		AckID:         ackID,
+	}, nil
+}
+
+// Decode the ackId parts
+func decode(parts []string) []string {
+	decodedParts := make([]string, 0)
+	for _, s := range parts {
+		decodedParts = append(decodedParts, utils.Decode(s))
 	}
+	return decodedParts
+}
+
+func isValidAckID(parts []string) bool {
+	// We can add more validations in future as required
+
+	// Validating the len of parts of AckMessage.
+	if len(parts) != 7 {
+		return false
+	}
+
+	return true
 }
 
 var currentHostIP string
--- a/internal/subscriber/model.go
+++ b/internal/subscriber/model.go
@@ -190,20 +190,17 @@ func ParseAckID(ackID string) (*AckMessage, error) {
 
 	partition, err := strconv.ParseInt(decodedParts[3], 10, 0)
 	if err != nil {
-		err = errors.WithMessage(err, fmt.Sprintf("error in parsing partition value %v to int", decodedParts[3]))
-		return nil, err
+		return nil, errors.WithMessage(err, fmt.Sprintf("error in parsing partition value %v to int", decodedParts[3]))
 	}
 
 	offset, err := strconv.ParseInt(decodedParts[4], 10, 0)
 	if err != nil {
-		err = errors.WithMessage(err, fmt.Sprintf("error in parsing offset value %v to int", decodedParts[4]))
-		return nil, err
+		return nil, errors.WithMessage(err, fmt.Sprintf("error in parsing offset value %v to int", decodedParts[4]))
 	}
 
 	deadline, err := strconv.ParseInt(decodedParts[5], 10, 0)
 	if err != nil {
-		err = errors.WithMessage(err, fmt.Sprintf("error in parsing deadline value %v to int", decodedParts[5]))
-		return nil, err
+		return nil, errors.WithMessage(err, fmt.Sprintf("error in parsing deadline value %v to int", decodedParts[5]))
 	}
 
 	return &AckMessage{
--- a/internal/subscriber/model.go
+++ b/internal/subscriber/model.go
@@ -80,10 +80,17 @@ type AckMessage struct {
 }
 
 var (
-	errIllegalPartitionValue = errors.New("partition value cannot be less than 0")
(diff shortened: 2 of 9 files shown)


internal/stream/stream.go
```diff
@@ -240,7 +240,11 @@ func (ps *PushStream) nack(ctx context.Context, message *metrov1.ReceivedMessage
 	logFields["ackId"] = message.AckId
 
 	logger.Ctx(ctx).Infow("worker: sending nack request to subscriber", "logFields", logFields)
-	ackReq := subscriber.ParseAckID(message.AckId)
+	ackReq, err := subscriber.ParseAckID(message.AckId)
+	if err != nil {
+		logger.Ctx(ctx).Errorf("worker: error in parsing ackId", "msg", err.Error(), "logFields", logFields)
```
shivasishdas, line 245:
```suggestion
		logger.Ctx(ctx).Errorf("worker: error in parsing ackId", "error", err.Error(), "logFields", logFields)
```

internal/subscriber/model.go
```diff
@@ -151,14 +152,29 @@ func (a *AckMessage) HasHitDeadline() bool {
 }
 
 // ParseAckID ...
-func ParseAckID(ackID string) *AckMessage {
+func ParseAckID(ackID string) (*AckMessage, error) {
 	// split the message and parse tokens
 	parts := strings.Split(ackID, ackIDSeparator)
+	if !isValidAckID(parts) {
+		return nil, merror.Newf(merror.InvalidArgument, "AckID received is not in expected format")
```
shivasishdas, line 159: store the error in a var and return that.

internal/subscriber/model.go
```diff
@@ -151,14 +152,29 @@ func (a *AckMessage) HasHitDeadline() bool {
...
+	}
+	partition, err := strconv.ParseInt(utils.Decode(parts[3]), 10, 0)
+	if err != nil {
+		fmt.Println(err)
```
shivasishdas, line 163: remove

internal/subscriber/model.go
```diff
@@ -151,14 +152,29 @@ func (a *AckMessage) HasHitDeadline() bool {
...
+	}
+
+	offset, err := strconv.ParseInt(utils.Decode(parts[4]), 10, 0)
 
-	// TODO : add validations
-	partition, _ := strconv.ParseInt(utils.Decode(parts[3]), 10, 0)
-	offset, _ := strconv.ParseInt(utils.Decode(parts[4]), 10, 0)
-	deadline, _ := strconv.ParseInt(utils.Decode(parts[5]), 10, 0)
+	if err != nil {
+		return nil, fmt.Errorf("error in parsing offset value %v to int", utils.Decode(parts[4]))
```
shivasishdas, line 170: decode once, store in var and re-use the same. Do for all parts.

internal/subscriber/model.go
```diff
@@ -151,14 +152,29 @@ func (a *AckMessage) HasHitDeadline() bool {
...
+	}
+
+	deadline, err := strconv.ParseInt(utils.Decode(parts[5]), 10, 0)
+
```
shivasishdas, line 174: nit: extra \n

internal/subscriber/model_test.go
```diff
@@ -0,0 +1,39 @@
...
+		Deadline:      100,
+		MessageID:     "test-message-id",
+		AckID:         "MS4yLjMuNA==_dGVzdC1zdWI=_dGVzdC10b3BpYw==_MA==_MA==_MTAw_dGVzdC1tZXNzYWdlLWlk",
+	}
+}
+
+func TestModel_ParseAckID(t *testing.T) {
+	expectedAckMessage := getDummyAcknowledgeModel()
+	ackID := "MS4yLjMuNA==_dGVzdC1zdWI=_dGVzdC10b3BpYw==_MA==_MA==_MTAw_dGVzdC1tZXNzYWdlLWlk"
+	ackMessage, _ := ParseAckID(ackID)
```
shivasishdas, line 26: dont ignore `err`. Assert it to `nil`

shivasishdas (changes requested):

shivasishdas:
@rutul25 Address this TODO as well?
https://github.com/razorpay/metro/blob/9dc31b2cdbab093d1b89991aa49b925c6eacc310/internal/subscriber/model.go#L86

internal/subscriber/model.go
razorao: Log the error at the origin. This way the caller need not explicitly log your errors everytime

service/web/subscriberserver_test.go
```diff
@@ -254,7 +254,7 @@ func TestSubscriberServer_Acknowledge(t *testing.T) {
 	ctx := context.Background()
 	req := &metrov1.AcknowledgeRequest{
 		Subscription: "projects/project321/subscriptions/testsub",
-		AckIds:       []string{"a_a_a_0_0_1_1"},
+		AckIds:       []string{"MS4yLjMuNA==_dGVzdC1zdWI=_dGVzdC10b3BpYw==_MA==_MA==_MTAw_dGVzdC1tZXNzYWdlLWlk"},
```
razorao, line 257: Since the ackid is being re-used in multiple tests, have this in one variable `validAckID` ?

razorao (changes requested):

internal/stream/stream.go
rutul25: Done

internal/subscriber/model.go
rutul25: So this function didn't have any context. So I have changed the signature and taking in context from caller method and logging the error in this function itself before passing it back to the caller. Do post any suggestions/changes in this approach

internal/subscriber/model.go
rutul25: Done

internal/subscriber/model.go
rutul25: Done

rutul25:
> @rutul25 Address this TODO as well?
> https://github.com/razorpay/metro/blob/9dc31b2cdbab093d1b89991aa49b925c6eacc310/internal/subscriber/model.go#L86

Added few primitive checks on the int fields. Do suggest any more checks if you had any on your mind.

service/web/subscriberserver_test.go
rutul25: Done

internal/subscriber/model.go
```diff
@@ -81,16 +84,32 @@ type AckMessage struct {
 const ackIDSeparator = "_"
 
 // NewAckMessage ...
-func NewAckMessage(subscriberID, topic string, partition, offset, deadline int32, messageID string) IAckMessage {
-	// TODO: add needed validations on all fields
+func NewAckMessage(subscriberID, topic string, partition, offset, deadline int32, messageID string) (IAckMessage, error) {
+
+	//Validating parameters for illegal values
+	if partition < 0 {
+		err := errors.New("partition value cannot be less than 0")
```
shivasishdas, line 91: move these error var's out of the function to the top of the file. You can then assert these in your UTs.

internal/subscriber/model.go
```diff
@@ -151,25 +170,68 @@ func (a *AckMessage) HasHitDeadline() bool {
...
+	if !isValidAckID(decodedParts) {
+		err := merror.Newf(merror.InvalidArgument, "AckID received is not in expected format")
+		logger.Ctx(ctx).Errorw(err.Error())
+		return nil, err
+	}
+
+	partition, err := strconv.ParseInt(decodedParts[3], 10, 0)
+	if err != nil {
+		logger.Ctx(ctx).Errorw(err.Error())
+		err = fmt.Errorf("error in parsing partition value %v to int", decodedParts[3])
```
shivasishdas, line 187: try using the below instead and see if this helps create a more descriptive error.
```
errors.WithMessage(err , fmt.Sprintf("error in parsing partition value %v to int", decodedParts[3]))
```

And no need to log here if the calling function is doing this. Can remove the duplicate log lines. In turn, can remove ctx as well. Do check the file for more such instances.

service/web/stream/request.go
```diff
@@ -57,14 +62,18 @@ func NewParsedStreamingPullRequest(req *metrov1.StreamingPullRequest) (*ParsedSt
 // NewParsedAcknowledgeRequest ...
 func NewParsedAcknowledgeRequest(req *metrov1.AcknowledgeRequest) (*ParsedStreamingPullRequest, error) {
 	parsedReq := &ParsedStreamingPullRequest{}
+	ctx := context.Background()
```
shivasishdas, line 65: nope. Use the calling functions context.

service/web/stream/request.go
```diff
@@ -77,6 +86,7 @@ func NewParsedAcknowledgeRequest(req *metrov1.AcknowledgeRequest) (*ParsedStream
 // NewParsedModifyAckDeadlineRequest ...
 func NewParsedModifyAckDeadlineRequest(req *metrov1.ModifyAckDeadlineRequest) (*ParsedStreamingPullRequest, error) {
 	parsedReq := &ParsedStreamingPullRequest{}
+	ctx := context.Background()
```
shivasishdas, line 89: same

internal/subscriber/model_test.go
```diff
@@ -0,0 +1,93 @@
...
+	assert.Nil(t, err)
+	assert.Equal(t, expectedAckMessage, ackMessage)
+}
+
+func TestModel_NewAckMessage_InvalidPartition(t *testing.T) {
+	ackMessage, err := NewAckMessage("test-sub", "test-topic", -1, 0, 100, "test-message-id")
+	if err == nil {
+		t.Errorf("Expected error but received %v", ackMessage)
+	}
+	expectedErr := errors.New("partition value cannot be less than 0")
```
shivasishdas, line 73: assert with the `errors` stored in `var` as per the above comments.

shivasishdas (changes requested):

internal/subscriber/model.go
rutul25: I can assert the errors in the test this way also right? As we are returning err from the function, I can use the return values to assert the returned error. Is there any added benefits of moving the err variable to the top of the file?

service/web/stream/request.go
rutul25: Remove this as no longer required after removing log statements from model file

internal/subscriber/model.go
rutul25: Done

internal/subscriber/model.go
razorao: Two things:
1. If you declare all the errors in the top of the file, it lets everyone know all the kind of errors that your code throws.
2. It looks a little neater (personal opinion)
3. Asserting the same error var in test cases ensures to a good extent that, any modifications in future w.r.t the same error gets reflected in test cases as well.. no need to change the test cases.

rutul25: Okay Got it. Is there any file in which this is followed that I can use as a reference?

razorao:
@rutul25 please rebase this branch with master

razorao (approved):
Most of my review comments are nit pickings. Approving the PR. But please do make the changes (if needed incrementally)

internal/subscriber/model.go
razorao: https://github.com/razorpay/metro/blob/a494477e00cd8b976e3e2569a0df527fecdcf317/internal/subscription/validation.go#L62

razorao (approved):

internal/subscriber/model.go
```diff
@@ -78,19 +79,39 @@ type AckMessage struct {
 	AckID         string
 }
 
+var (
+	errIllegalPartitionValue = errors.New("partition value cannot be less than 0")
```
shivasishdas, line 83: export these

internal/subscriber/model.go
```diff
@@ -151,25 +172,63 @@ func (a *AckMessage) HasHitDeadline() bool {
...
-	partition, _ := strconv.ParseInt(utils.Decode(parts[3]), 10, 0)
-	offset, _ := strconv.ParseInt(utils.Decode(parts[4]), 10, 0)
-	deadline, _ := strconv.ParseInt(utils.Decode(parts[5]), 10, 0)
+	if !isValidAckID(decodedParts) {
+		return nil, errInvalidAckID
+	}
+
+	partition, err := strconv.ParseInt(decodedParts[3], 10, 0)
+	if err != nil {
+		err = errors.WithMessage(err, fmt.Sprintf("error in parsing partition value %v to int", decodedParts[3]))
```
shivasishdas, line 186:
```suggestion
		return nil, errors.WithMessage(err, fmt.Sprintf("error in parsing partition value %v to int", decodedParts[3]))
```
nit: could have done away with the redundant line to init the error 👀 what say?

internal/subscriber/model.go
```diff
@@ -151,25 +172,63 @@ func (a *AckMessage) HasHitDeadline() bool {
...
 		Offset:        int32(offset),
 		Deadline:      int32(deadline),
-		MessageID:     utils.Decode(parts[6]),
+		MessageID:     decodedParts[6],
 		AckID:         ackID,
+	}, nil
+}
+
+// Decode the ackId parts
+func decode(parts []string) []string {
```
shivasishdas, line 215: move to `pkg/utils`? Technically this `func` just takes a slice of encoded strings and returns a slice of decoded strings. Nothing specific to do with this `pkg`

internal/subscriber/model_test.go
```diff
@@ -0,0 +1,86 @@
...
+	ackMessage, err := ParseAckID(ackID)
+	assert.Nil(t, err)
+	assert.Equal(t, expectedAckMessage, ackMessage)
+}
+
+func TestModel_ParseAckID_InvalidInput(t *testing.T) {
+	ackID := inValidAckID
+	ackMessage, err := ParseAckID(ackID)
+	expectedErr := merror.Newf(merror.InvalidArgument, "AckID received is not in expected format")
+	if err == nil {
```
shivasishdas, line 50: not needed right? the assertion itself would fail and log the error.

internal/subscriber/model_test.go
```diff
@@ -0,0 +1,86 @@
...
+func TestModel_NewAckMessage_ValidInput(t *testing.T) {
+	ackMessage, err := NewAckMessage("test-sub", "test-topic", 0, 0, 100, "test-message-id")
+	expectedAckMessage := getValidAcknowledgeMessage()
+	assert.Nil(t, err)
+	assert.Equal(t, expectedAckMessage, ackMessage)
+}
+
+func TestModel_NewAckMessage_InvalidPartition(t *testing.T) {
+	ackMessage, err := NewAckMessage("test-sub", "test-topic", -1, 0, 100, "test-message-id")
+	if err == nil {
```
shivasishdas, line 66: same as above

internal/subscriber/model_test.go
```diff
@@ -0,0 +1,86 @@
...
+		t.Errorf("Expected error but received %v", ackMessage)
+	}
+	assert.Equal(t, errIllegalPartitionValue, err)
+}
+
+func TestModel_NewAckMessage_InvalidOffset(t *testing.T) {
+	ackMessage, err := NewAckMessage("test-sub", "test-topic", 0, -5, 100, "test-message-id")
+	if err == nil {
```
shivasishdas, line 74: same as above

internal/subscriber/model_test.go
```diff
@@ -0,0 +1,86 @@
...
+		t.Errorf("Expected error but received %v", ackMessage)
+	}
+	assert.Equal(t, errIllegalOffsetValue, err)
+}
+
+func TestModel_NewAckMessage_InvalidDeadline(t *testing.T) {
+	ackMessage, err := NewAckMessage("test-sub", "test-topic", 0, 0, -100, "test-message-id")
+	if err == nil {
```
shivasishdas, line 82: same as above

shivasishdas (changes requested):

internal/subscriber/model.go
rutul25: I think in one of the previous comments, you suggested to take out the error and initialise it in a var and use that

[https://github.com/razorpay/metro/pull/242#discussion_r703400310](url)

internal/subscriber/model.go
rutul25: make sense. Will move it to utils

internal/subscriber/model.go
rutul25: Done

pkg/utils/utils.go
```diff
@@ -12,3 +12,12 @@ func Decode(input string) string {
 	decoded, _ := base64.StdEncoding.DecodeString(input)
 	return string(decoded)
 }
+
+// DecodeSlice given a slice of base64 encoded strings, returns a slice of decoded strings
+func DecodeSlice(input []string) []string {
```
shivasishdas, line 17: add UT as well for this

internal/subscriber/model.go
shivasishdas: That was when you were logging the same error as well. Then init once and re-using the `err var` in the `logline` and in the `return` made sense. Now that we have removed the logline we can simply return.

shivasishdas (changes requested):

internal/subscriber/model.go
shivasishdas: My previous comment got deleted somehow. Re-posting again.

Previously we were logging and returning the same err hence keeping in a var made sense to avoid duplicates. Now we have removed the logger so a direct return makes sense.

shivasishdas (approved):
LGTM

PR #248, opened 2021-09-10
Retry PR Fixes:

razorao:
- Added max >= min validation
 - fixed the nil retrier error


internal/subscription/delay.go
```diff
@@ -94,6 +96,10 @@ func validateDelayConfig(model *Model) error {
 		}
 	}
 
+	if minB > maxB {
+		return ErrInvalidMinAndMaxBackoff
```
shivasishdas, line 100: Let's add a UT for this as well

shivasishdas (changes requested):

sunnyagg (approved):

PR #252, opened 2021-09-14
adds support for editing retry/dlq configs in subscription

sunnyagg:
- Fixes/Refactor validations
- Adds support for updating retry/dlq configs


internal/subscription/validation.go
```diff
@@ -114,78 +155,206 @@ func GetValidatedModelForUpdate(ctx context.Context, req *metrov1.Subscription)
...
+			}
+
+			if p, ok := pushAttr[attributePassword]; ok {
+				if strings.Trim(p, " ") == "" {
+					return ErrInvalidPushEndpointPassword
+				}
+				password = p
+			}
+
+			delete(pushAttr, attributeUsername)
```
sunnyagg, line 245: @rutul25 this should partial fix the creds being present in attributes, you will still need to have a way to fix the exisiting data models.

internal/subscription/validation.go
```diff
@@ -23,8 +25,68 @@ const (
...
-	pushConfigPath     = "push_config"
-	ackDeadlineSecPath = "ack_deadline_seconds"
+	pushConfigPath       = "push_config"
+	ackDeadlineSecPath   = "ack_deadline_seconds"
+	retryConfigPath      = "retry_policy"
+	deadLetterConfigPath = "dead_letter_policy"
+)
+
+// Config validations related values
+const (
```
razorao, line 35: the consts and the error vars defined here seem to be relavent only inside the subscription package. Do you think it would be better if we make these constants private instead of public? i.e names starting with small case.

internal/subscription/validation.go
```diff
@@ -114,78 +155,206 @@ func GetValidatedModelForUpdate(ctx context.Context, req *metrov1.Subscription)
...
+			PushEndpoint: urlEndpoint,
+			Attributes:   pushAttr,
+			Credentials:  creds,
+		}
+	}
+
+	return nil
+}
+
+func getValidatedAckDeadline(_ context.Context, m *Model, req *metrov1.Subscription) error {
```
razorao, line 264: Just like every other validation function you have written here, this model does not return a validated value but modifies the model. Therefore, perhaps the name `validateAckDeadline` is more apt here. Thoughts?

internal/subscription/validation.go
```diff
@@ -114,78 +155,206 @@ func GetValidatedModelForUpdate(ctx context.Context, req *metrov1.Subscription)
...
+
+	m.RetryPolicy = &RetryPolicy{
+		MinimumBackoff: uint(retryPolicy.MinimumBackoff.Seconds),
+		MaximumBackoff: uint(retryPolicy.MaximumBackoff.Seconds),
+	}
+
+	return nil
+}
+
+func validatedDeadLetterPolicy(_ context.Context, m *Model, req *metrov1.Subscription) error {
```
razorao, line 318: nit: `validateDeadLetterPolicy`

internal/subscription/validation.go
```diff
@@ -114,78 +155,206 @@ func GetValidatedModelForUpdate(ctx context.Context, req *metrov1.Subscription)
...
 }
 
 func extractSubscriptionMetaAndValidate(_ context.Context, name string) (projectID string, subscriptionName string, err error) {
 	match := subscriptionNameRegex.FindStringSubmatch(name)
 	if len(match) != 3 {
-		err = fmt.Errorf("invalid subscription name")
-		return "", "", err
+		return "", "", errors.New(fmt.Sprintf("Invalid [subscriptions] name: (name=%s)", name))
 	}
 	projectID = subscriptionNameRegex.FindStringSubmatch(name)[1]
```
razorao, line 354: projectID = match[1]
subscriptionName = match[2]
This would make things more efficient, wouldn't it?

razorao (changes requested):

internal/subscription/validation.go
sunnyagg: will change the name, missed this one

internal/subscription/validation.go
sunnyagg: Yes, we dont need to parse multiple times

internal/subscription/validation.go
sunnyagg: if we are writing functional tests outside of package it might be usefule to use the same constants in tests while testing. 
Today our errors are not 100% compatible with google pubsub, we will need to relook at all our error codes. can take this up then.

razorao (approved):

PR #257, opened 2021-09-29
Updated Sub Migration API to remove unencrypted push creds

rutul25:
- Updated the subscription migration API to remove the unencrypted push credentials that are stored in the attribute field of push config as a part of the old sub model


internal/subscription/core.go
```diff
@@ -343,54 +342,39 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
...
-		// update retry policy if not set
-		if model.RetryPolicy == nil {
-			model.setDefaultRetryPolicy()
-			// update dead letter policy every time as previously we were saving only dead-letter topic names
-			// this takes care of updating max_delivery_attempts
-			model.setDefaultDeadLetterPolicy()
-			needsUpdate = true
+	updatedSubCount := 0
+	for _, sub := range subscriptionsToUpdate {
+		if sub.PushConfig == nil {
```
shivasishdas, line 347:
```suggestion
		if sub.PushConfig == nil || sub.PushConfig.Attributes == nil {
```
no need for `if pushAttr != nil` check below.

internal/subscription/core.go
```diff
@@ -343,54 +342,39 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
...
-		tModel, terr := topic.GetValidatedModel(ctx, &metrov1.Topic{
-			Name: tName,
-		})
-		if terr != nil {
-			logger.Ctx(ctx).Errorw("migration: failed to create validated topic model", "tName", tName, "error", terr.Error())
-			failedTopicNames = append(failedTopicNames, tName)
-			continue
-		}
+			if _, ok := pushAttr[attributePassword]; ok {
+				logger.Ctx(ctx).Infow("migration: deleting password", "password", pushAttr[attributePassword], "subscription", sub.Name)
```
shivasishdas, line 360: Don't log. `password` is sensitive info. The log can say `deleting password for username` instead

internal/subscription/core.go
```diff
@@ -343,54 +342,39 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
...
 
-		err := c.topicCore.CreateTopic(ctx, tModel)
-		if err != nil {
-			logger.Ctx(ctx).Errorw("migration: failed to create delay topic", "tName", tName, "error", err.Error())
-			failedTopicNames = append(failedTopicNames, tName)
-			continue
+			if needsUpdate {
+				updatedSubCount++
+				logger.Ctx(ctx).Infow("migration: updating subscription", "subscription", sub.Name)
+				sub.PushConfig.Attributes = pushAttr
```
shivasishdas, line 368: don't think this is needed as it is using the same reference for delete. Do verify.

internal/subscription/core.go
```diff
@@ -343,54 +342,39 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
...
+					logger.Ctx(ctx).Errorw("migration: error in updating subscription", "subscription", sub.Name, "error", err.Error())
+					return err
+				}
+			}
 		}
-		successTopicNames = append(successTopicNames, tName)
 	}
 
-	logger.Ctx(ctx).Infow("migration: request completed", "successTopicNames", successTopicNames, "failedTopicNames", failedTopicNames)
+	logger.Ctx(ctx).Infow("migration: request completed.", "Subscriptions Updated", updatedSubCount)
```
shivasishdas, line 378:
```suggestion
	logger.Ctx(ctx).Infow("migration: request completed.", "subscriptionsUpdated", updatedSubCount)
```
keep this as one word, easier to do an exact match on Sumo.

internal/subscription/core.go
```diff
@@ -343,54 +342,39 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
...
-			topicNames = append(topicNames, model.GetDelayTopics()...)
-
-			logger.Ctx(ctx).Infow("migration: updating subscription", "model", model.Name)
-			// update the subscription model
-			err := c.UpdateSubscription(ctx, model)
-			if err != nil {
-				return err
+		pushAttr := sub.PushConfig.Attributes
+		if pushAttr != nil {
+			needsUpdate := false
```
shivasishdas, line 352: nit: move line right after loop begin.

shivasishdas (changes requested):

shivasishdas (approved):
LGTM. Do test on non-prod once before merging.

rutul25:
> LGTM. Do test on non-prod once before merging.

Sure. Will do migration testing on stage first

sunnyagg (approved):

razorao (approved):
@rutul25 Just one thought: In case something goes wrong, maybe it will help if we get the error(s) as response for the API. Else we will have to depend on logs.
Just a thought, not a recommendation

internal/subscription/core.go
```diff
@@ -343,54 +342,36 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
 
 	logger.Ctx(ctx).Infow("migration: found subscriptions to migrate", "count", len(subscriptionsToUpdate))
 
-	topicNames := make([]string, 0)
-	for _, model := range subscriptionsToUpdate {
+	updatedSubCount := 0
+	for _, sub := range subscriptionsToUpdate {
```
vnktram, line 346: While acting on all objects of an entity always batch the operation. This seems harmless now but down the line a migration can run for multiple hours and consume a large memory footprint (Need not change as part of this PR)

vnktram (approved):
LGTM

PR #259, opened 2021-10-06
Fix issues with span propagation

vnktram:
- [ ] Fixes issues with trace injection at Producer end
- [ ] Introduces span propagation via downstream HTTP client (Push Subscriber)
- [ ] Introduces gRPC server interceptors for Streaming requests (Pull Subscriber)


pkg/messagebroker/kafka.go
```diff
@@ -392,7 +392,7 @@ func (k *KafkaBroker) SendMessage(ctx context.Context, request SendMessageToTopi
 		TopicPartition: kafkapkg.TopicPartition{Topic: &topicN, Partition: kafkapkg.PartitionAny},
 		Value:          request.Message,
 		Key:            []byte(request.OrderingKey),
-		Headers:        kHeaders,
+		Headers:        carrier,
```
sunnyagg, line 395: isn't carrier updating the same set of headers ?

vnktram: Yes but we pass the kHeaders object to carrier and not it's reference. Hence the original object does not get updated.

PR #260, opened 2021-10-07
Ordering Prep

razorao:
- Messages were returned in a map when receiving from a consumer. This did not serve any purpose. Moreover, we loose the order of messages when iterating over a map. Hence, made changes to return the messages as an array/slice.
- Subscriber was managing the states of the consumer. With ordering, the subscriber would have to manage consumption states from 2 topics (primary and retry pause/resume). Moved the entire logic of managing consumers into a consumer manager.


vnktram (approved):
LGTM

PR #261, opened 2021-10-08
Issue #221:  Added Admin APIs for Fetching and Listing Project Credentials

### LINKED ISSUE #221: Admin route to fetch/verify project creds
Currently we don't have route to verify the creds or fetch the lost creds. we can add admin routes for same

rutul25:
- Added Fetch ProjectCredentials API in admin route which can be used to get lost credentials using projectId and username.
- Also added List ProjectCredentials API which will list out all the credentials for the given projectId. Note that password in this output will be masked.

### DIFF
--- a/internal/credentials/core_test.go
+++ b/internal/credentials/core_test.go
@@ -76,7 +76,7 @@ func TestCore_List_Success(t *testing.T) {
 	ctx := context.Background()
 
 	project := "project123"
-	prefix := common.GetBasePrefix() + Prefix + project
+	prefix := common.GetBasePrefix() + Prefix + project + "/"
 
 	var model common.IModel = &Model{
 		Username:  "user1",
@@ -106,7 +106,7 @@ func TestCore_List_Failure(t *testing.T) {
 	ctx := context.Background()
 
 	project := "project123"
-	prefix := common.GetBasePrefix() + Prefix + project
+	prefix := common.GetBasePrefix() + Prefix + project + "/"
 
 	mockRepo.EXPECT().List(ctx, prefix).Return(nil, errors.New("Error getting credentials"))
 
--- a/internal/credentials/core_test.go
+++ b/internal/credentials/core_test.go
@@ -66,3 +66,19 @@ func TestCore_Get(t *testing.T) {
 	_, err := c.Get(ctx, project, username)
 	assert.NoError(t, err)
 }
+
+func TestCore_List(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mockRepo := mocks1.NewMockIRepo(ctrl)
+	mockCore := mocks2.NewMockICore(ctrl)
+	c := NewCore(mockRepo, mockCore)
+	ctx := context.Background()
+
+	project := "project123"
+	prefix := common.GetBasePrefix() + Prefix + project
+
+	mockRepo.EXPECT().List(ctx, prefix).Return(nil, nil)
+
+	_, err := c.List(ctx, project)
+	assert.NoError(t, err)
+}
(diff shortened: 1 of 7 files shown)


internal/credentials/core_test.go
```diff
@@ -66,3 +66,19 @@ func TestCore_Get(t *testing.T) {
...
+	ctrl := gomock.NewController(t)
+	mockRepo := mocks1.NewMockIRepo(ctrl)
+	mockCore := mocks2.NewMockICore(ctrl)
+	c := NewCore(mockRepo, mockCore)
+	ctx := context.Background()
+
+	project := "project123"
+	prefix := common.GetBasePrefix() + Prefix + project
+
+	mockRepo.EXPECT().List(ctx, prefix).Return(nil, nil)
```
razorao, line 80: returning nil, nil doesn't seem to be a valid scenario.
Would suggest 2 test cases:
1) Expect valid prefix and return valid list with no error
2) Expect valid prefix but return error

internal/credentials/model.go
```diff
@@ -81,6 +84,14 @@ func (m *Model) GetPassword() string {
 	return string(pwd)
 }
 
+// GetHiddenPassword returns the Password after
+// masking all but last 4 characters with *
+func (m *Model) GetHiddenPassword() string {
+	// Returns a hidden password which contains a string of asterisk
+	// followed by last 4 characters of the original password
+	return AsteriskString + m.GetPassword()[len(m.GetPassword())-4:len(m.GetPassword())]
```
razorao, line 92: call `m.GetPassword()` once and store it in a variable. use the variable for subsequent index calculation/length calculation.

every call to `m.GetPassword()` decrypts from hex.

service/web/adminserver_test.go
```diff
@@ -153,3 +154,91 @@ func TestAdminServer_DeleteProjectValidationFailure(t *testing.T) {
...
+		Password: "p",
+	}
+
+	credential := &credentials.Model{
+		ProjectID: "test-project",
+		Username:  "test-project__00f790",
+		Password:  pwd,
+	}
+
+	exectedCredProto := &metrov1.ProjectCredentials{
```
razorao, line 184: nit: expectedCredProto

service/web/adminserver_test.go
```diff
@@ -153,3 +154,91 @@ func TestAdminServer_DeleteProjectValidationFailure(t *testing.T) {
...
+		Password: "p",
+	}
+
+	credential := []*credentials.Model{{
+		ProjectID: "test-project",
+		Username:  "test-project__00f790",
+		Password:  pwd,
+	}}
+
+	exectedCredProto := &metrov1.ProjectCredentialsList{
```
razorao, line 227: nit: expectedCredProto

internal/credentials/model.go
razorao: Also, check if the length of `m.GetPassword` is > 4 ? probably there is no chance that the password is shorter than a 4 letter character.. At least mention this as a warning?

razorao (changes requested):

internal/credentials/model.go
rutul25: I think we should add some validations in `func NewCredential(username, password string) *Model`  to make sure valid inputs are sent. For now, only len of password can be checked, but it can be extended in future if needed. There is a `IsValidUsername` which can be called. Thoughts?

razorao: Validation may not be required now since we are generating the password and not taking an input.

razorao (approved):

internal/credentials/model.go
rutul25: Okay. Adding a warning in `type Model struct` to mention that password len must be > 4

razorao (approved):

PR #268, opened 2021-10-28
Added filter support in create subscription

rutul25:
- This PR contains the changes for supporting the filtering capability while creating the subscription.
- Using filter expression, subscribers can filter messages by their attributes. 
- Separate PR/Revision will be raised which will include the logic of using the filter details to evaluate the incoming messages and filter out relevant messages to be sent to the subscriber. 
- For each message that is sent to the topic, it's message attributes will be used to call the evaluation function for the subscription filter and will be sent to the subscriber only if matches the requirements.

### DIFF
--- a/internal/subscription/model.go
+++ b/internal/subscription/model.go
@@ -27,7 +27,7 @@ type Model struct {
 	EnableMessageOrdering          bool              `json:"enable_message_ordering,omitempty"`
 	ExpirationPolicy               *ExpirationPolicy `json:"expiration_policy,omitempty"`
 	FilterExpression               string            `json:"filter,omitempty"`
-	FilterStruct                   *Filter           `json:"filter_struct,omitempty"`
+	filterStruct                   *Filter
 	RetryPolicy                    *RetryPolicy      `json:"retry_policy,omitempty"`
 	DeadLetterPolicy               *DeadLetterPolicy `json:"dead_letter_policy,omitempty"`
 	Detached                       bool              `json:"detached,omitempty"`
@@ -203,4 +203,18 @@ func (m *Model) GetDelayConsumerGroupID(delayTopic string) string {
 // GetDelayConsumerGroupInstanceID returns the consumer group ID to be used by the specific delay consumer
 func (m *Model) GetDelayConsumerGroupInstanceID(subscriberID, delayTopic string) string {
 	return fmt.Sprintf(delayConsumerGroupInstanceIDFormat, delayTopic, subscriberID)
+}
+
+// GetFilterExpressionAsStruct parses and returns the filter expression into GO Struct
+func (m *Model) GetFilterExpressionAsStruct() (*Filter, error) {
+	if m.filterStruct != nil {
+		return m.filterStruct, nil
+	}
+	f := &Filter{}
+	err := filter.Parser.ParseString("", m.FilterExpression, f)
+	if err != nil {
+		return nil, err
+	}
+	m.filterStruct = f
+	return f, nil
 }
--- a/internal/subscription/model.go
+++ b/internal/subscription/model.go
@@ -6,6 +6,7 @@ import (
 	"github.com/razorpay/metro/internal/common"
 	"github.com/razorpay/metro/internal/credentials"
 	"github.com/razorpay/metro/internal/topic"
+	filter "github.com/razorpay/metro/pkg/filtering"
 )
 
 const (
@@ -25,7 +26,8 @@ type Model struct {
 	Labels                         map[string]string `json:"labels,omitempty"`
 	EnableMessageOrdering          bool              `json:"enable_message_ordering,omitempty"`
 	ExpirationPolicy               *ExpirationPolicy `json:"expiration_policy,omitempty"`
-	Filter                         string            `json:"filter,omitempty"`
+	FilterExpression               string            `json:"filter,omitempty"`
+	FilterStruct                   *Filter           `json:"filter_struct,omitempty"`
 	RetryPolicy                    *RetryPolicy      `json:"retry_policy,omitempty"`
 	DeadLetterPolicy               *DeadLetterPolicy `json:"dead_letter_policy,omitempty"`
 	Detached                       bool              `json:"detached,omitempty"`
@@ -34,6 +36,9 @@ type Model struct {
 	ExtractedTopicName             string            `json:"extracted_topic_name"`
 	ExtractedSubscriptionName      string            `json:"extracted_subscription_name"`
 }
+
+// Filter defines the Filter criteria for messages
+type Filter = filter.Condition
 
 // PushConfig defines the push endpoint
 type PushConfig struct {
--- a/internal/subscription/validation.go
+++ b/internal/subscription/validation.go
@@ -12,7 +12,6 @@ import (
 	"github.com/razorpay/metro/internal/credentials"
 	"github.com/razorpay/metro/internal/merror"
 	"github.com/razorpay/metro/internal/topic"
-	filter "github.com/razorpay/metro/pkg/filtering"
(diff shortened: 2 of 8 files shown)


pkg/filtering/evaluate.go
```diff
@@ -0,0 +1,138 @@
...
+func (e *basicExpression) Evaluate(attrs map[string]string) (bool, error) {
+	switch {
+	case e.Has != nil:
+		return e.Has.Evaluate(attrs)
+	case e.Value != nil:
+		return e.Value.Evaluate(attrs)
+	case e.Predicate != nil:
+		return e.Predicate.Evaluate(attrs)
+	default:
+		return false, errors.New("unpopulated BasicExpression")
```
vnktram, line 43: Please follow the log message standard we have across the code base. (component: Error occured) and contextual fields.

pkg/filtering/evaluate.go
```diff
@@ -0,0 +1,138 @@
...
+	if !ok {
+		return false, nil
+	}
+	switch e.Op {
+	case opEqual:
+		return v == e.Value, nil
+	case opNotEqual:
+		return v != e.Value, nil
+	default:
+		return false, fmt.Errorf("invalid Op '%s'", e.Op)
```
vnktram, line 98: nit: log format

pkg/filtering/evaluate.go
```diff
@@ -0,0 +1,138 @@
...
+func (e *hasAttributePredicate) Evaluate(attrs map[string]string) (bool, error) {
+	v, ok := attrs[e.Name]
+	if !ok {
+		return false, nil
+	}
+	switch e.Predicate {
+	case predicateHasPrefix:
+		return strings.HasPrefix(v, e.Value), nil
+	default:
+		return false, fmt.Errorf("invalid predicate '%s'", e.Predicate)
```
vnktram, line 112: nit: log format

pkg/filtering/evaluate.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+func andTerms(attrs map[string]string, terms []*term) (result bool, err error) {
+	if len(terms) == 0 {
+		return false, errors.New("AND requires a non-empty term list")
```
vnktram, line 118: Please provide contextual data on which filter resulted in this error.

pkg/filtering/evaluate.go
```diff
@@ -0,0 +1,138 @@
...
+		if result, err = t.Evaluate(attrs); err != nil || !result {
+			return
+		}
+	}
+	return
+}
+
+func orTerms(attrs map[string]string, terms []*term) (result bool, err error) {
+	if len(terms) == 0 {
+		return false, errors.New("OR requires a non-empty term list")
```
vnktram, line 130: Refer previous comment.

pkg/filtering/parser.go
```diff
@@ -0,0 +1,36 @@
...
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+package filter
+
+import (
+	"github.com/alecthomas/participle/v2"
+)
+
+var defaultParserOptions = []participle.Option{
+	participle.UseLookahead(50),
```
vnktram, line 27: What does this 50 signify? Can we add a comment on why we use 50

vnktram (changes requested):

pkg/filtering/grammar.go
```diff
@@ -0,0 +1,75 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+// FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+// COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+// IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+package filter
+
```
razorao, line 21: would suggest adding a few examples of filter expressions as comments so that we would be able to relate to the below structs while reading them.

internal/subscription/validation.go
```diff
@@ -185,6 +192,22 @@ func validateSubscriptionName(ctx context.Context, m *Model, req *metrov1.Subscr
 	return nil
 }
 
+func validateFilterExpression(ctx context.Context, m *Model, req *metrov1.Subscription) error {
+	if req.Filter != "" {
+		// validate that the filter expression is in the expected format.
+		// Also convert the expression into a GO Struct
+		f := &Filter{}
+		err := filter.Parser.ParseString(m.Name, req.Filter, f)
```
razorao, line 200: what's the role of passing `m.Name` here?

razorao:
@rutul25 Slightly confused. Since this is a PR where we only validate filter but don't actually evaluate it, maybe you can update the PR description on how you plan to use the code to run evaluate in the subsequent PR?

razorao (commented):
UTs are missing. Planning for a separate PR?

pkg/filtering/evaluate.go
razorao: Declare the error beforehand just like we have done it elsewhere.

rutul25: Done

pkg/filtering/evaluate.go
rutul25: Done

pkg/filtering/grammar.go
rutul25: Added few examples as suggested

rutul25:
> @rutul25 Slightly confused. Since this is a PR where we only validate filter but don't actually evaluate it, maybe you can update the PR description on how you plan to use the code to run evaluate in the subsequent PR?

Updated PR description with briefing on evaluation logic

rutul25:
> UTs are missing. Planning for a separate PR?

Yep. Will add separately after the evaluation PR

internal/subscription/validation.go
rutul25: Parser expects fileName in this argument. Given that we use string as an input and fileName is not relevant in our use case, have used subscription name instead for context.

razorao: But what's the impact? Does it create a file locally? Does it read a file? If it creates file(s) then we need to think of space usage/cleanup.

internal/subscription/model.go
```diff
@@ -25,7 +26,8 @@ type Model struct {
 	Labels                         map[string]string `json:"labels,omitempty"`
 	EnableMessageOrdering          bool              `json:"enable_message_ordering,omitempty"`
 	ExpirationPolicy               *ExpirationPolicy `json:"expiration_policy,omitempty"`
-	Filter                         string            `json:"filter,omitempty"`
+	FilterExpression               string            `json:"filter,omitempty"`
+	FilterStruct                   *Filter           `json:"filter_struct,omitempty"`
```
sunnyagg, line 30: do we need to save struct to db ? should we just save expression to db and have in memory struct 
can we just have that as a private attribute with getters ?

rutul25: Removed the struct from db model. Instead keeping it a private field with a Getter as suggested.

sunnyagg (approved):

vnktram (approved):

razorao (approved):

PR #269, opened 2021-10-29
Add offset management on Ack

vnktram:
- [x] Store latest committed offset to registry.
- [x] Rollback latest commit if broker commit fails
- [x] Add support for persisting ordering key offset.
- [x] Support delete operation on offset

### DIFF
--- a/internal/offset/core.go
+++ b/internal/offset/core.go
@@ -44,11 +44,11 @@ func (c *Core) SetOffset(ctx context.Context, m *Model) error {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "OffsetCore.CreateOffset")
 	defer span.Finish()
 
-	OffsetOperationCount.WithLabelValues(env, "SetOffset").Inc()
+	offsetOperationCount.WithLabelValues(env, "SetOffset").Inc()
 
 	startTime := time.Now()
 	defer func() {
-		OffsetOperationTimeTaken.WithLabelValues(env, "SetOffset").Observe(time.Now().Sub(startTime).Seconds())
+		offsetOperationTimeTaken.WithLabelValues(env, "SetOffset").Observe(time.Now().Sub(startTime).Seconds())
 	}()
 	if m.LatestOffset == "" {
 		return merror.Newf(merror.NotFound, "No offset provided for key %s", m.Key())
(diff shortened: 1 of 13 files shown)


razorao (changes requested):

internal/offset/core.go
```diff
@@ -0,0 +1,159 @@
...
+	if !ok {
+		return c.repo.Save(ctx, m)
+	}
+	// Extract offset to avoid being overriden
+	newOffsetToSet := m.LatestOffset
+	m, err = c.GetOffset(ctx, m)
+	if err != nil {
+		return merror.Newf(merror.NotFound, "failed to update offset")
+	}
+	rollbackOffset := m.rollbackOffset
```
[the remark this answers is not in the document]
vnktram: This has been described in the block comment at the top of the page. I'll add a comment here as well.

internal/subscriber/subscriber.go
```diff
@@ -207,7 +210,21 @@ func (s *Subscriber) acknowledge(req *AckMessage) {
 	}
 
 	if shouldCommit {
-		_, err := s.consumer.CommitByPartitionAndOffset(ctx, messagebroker.CommitOnTopicRequest{
+		offsetUpdated := true
+		registryOffset := strconv.Itoa(int(offsetToCommit) + 1)
+		offsetModel := offset.Model{
```
razorao, line 215: don't we need to do the same thing in `subscriber.retry` when we are comitting on the primary topic?

internal/subscriber/subscriber.go
```diff
@@ -207,7 +210,21 @@ func (s *Subscriber) acknowledge(req *AckMessage) {
...
+			Topic:        s.subscription.Topic,
+			Subscription: s.subscription.ExtractedSubscriptionName,
+			Partition:    req.Partition,
+			LatestOffset: registryOffset,
+		}
+		err := s.offsetCore.SetOffset(ctx, &offsetModel)
```
razorao, line 221: don't we need to do the same thing also in `subscriber.retry` when we are comitting on the primary topic?

razorao (approved):

internal/subscriber/subscriber.go
vnktram: Currently retry topics are not covered since retry might be in an ambiguous state when we rollback. The idea is to resume from the last successful ACK and retry all messages including ones pushed to the retry topics.

razorao (approved):

PR #273, opened 2021-11-10
Ordering/validation

razorao:
- Validate the publish request
- Test cases for validation

### DIFF
--- a/internal/publisher/core.go
+++ b/internal/publisher/core.go
@@ -2,6 +2,7 @@ package publisher
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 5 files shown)


internal/publisher/core.go
```diff
@@ -43,6 +44,15 @@ func (p *Core) Publish(ctx context.Context, req *metrov1.PublishRequest) ([]stri
 
 	msgIDs := make([]string, 0)
 
+	orderingKey := ""
+	if len(req.Messages) > 0 {
+		orderingKey = req.Messages[0].OrderingKey
```
sunnyagg, line 49: // All messages in request will have the same ordering key 

^ Didn't get this restriction, why we want to put this restriction ?

sunnyagg (approved):

Issue #274, opened 2021-11-10
Configurable http timeouts for push subscription


razorao:
Currently, while subscription's `ackDeadlineSecs` is intended to be used as the timeout value for push client, only the TCP connection timeout is being set. Hence in cases where TCP connection is established but the server application delay's the response, the default timeout value of 25 seconds is being honored. 

Need to also ensure that the `ResponseHeaderTimeout` is also set per subscription

PR #275, opened 2021-11-10
#274: Fixed the timeout config bug and added subscriber start log

razorao:
- Setting the `ResponseHeaderTimeout` from `ackDeadlineSecs` of the subscription
 - Added log to mark start of subscriber

### DIFF
--- a/internal/stream/stream.go
+++ b/internal/stream/stream.go
@@ -317,7 +317,20 @@ func NewPushStream(ctx context.Context, nodeID string, subName string, subscript
 	// set http connection timeout from the subscription
 	if subModel.AckDeadlineSeconds != 0 {
 		// make sure to convert sec to milli-sec
-		config.ConnectTimeoutMS = int(subModel.AckDeadlineSeconds) * 1e3
+		// set the timeout value only if greater than the default
+		ackDeadlineMs := int(subModel.AckDeadlineSeconds) * 1000
+		if ackDeadlineMs > config.ConnectTimeoutMS {
+			config.ConnectTimeoutMS = ackDeadlineMs
+		}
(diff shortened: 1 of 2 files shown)


internal/stream/stream.go
```diff
@@ -317,7 +317,20 @@ func NewPushStream(ctx context.Context, nodeID string, subName string, subscript
 	// set http connection timeout from the subscription
 	if subModel.AckDeadlineSeconds != 0 {
 		// make sure to convert sec to milli-sec
-		config.ConnectTimeoutMS = int(subModel.AckDeadlineSeconds) * 1e3
+		// set the timeout value only if greater than the default
+		ackDeadlineMs := int(subModel.AckDeadlineSeconds) * 1000
+		if ackDeadlineMs > config.ConnectTimeoutMS {
+			config.ConnectTimeoutMS = ackDeadlineMs
```
vnktram, line 323: Do we intend to set both values here?

vnktram (changes requested):

internal/stream/stream.go
razorao: One is a TCP connection establishment timeout and the other is the HTTP timeout.
We will need both of them. Ideally TCP connection establishment won't take much time but in case if it does, we should be able to handle it as well.

vnktram: TCP connection establishment timeout should not be considered from AckDeadline right? AckDeadline is specific to per message delivery.

razorao: So this comes into picture when we publish to http endpoint. The default is 2s but since we were updating it with subscription deadline ack secs (which is 10s), not setting this would mean we are reducing it.. Kind of risky

vnktram (approved):

PR #276, opened 2021-11-12
Ingestion prep:  subscription backfill endpoint

vnktram:
- [x] Modifies the migration endpoint to create the subscription broker topics.
- [x] Create subscription broker topics when a new subscription is created.

### DIFF
--- a/internal/subscription/model.go
+++ b/internal/subscription/model.go
@@ -81,6 +81,11 @@ func (m *Model) GetTopic() string {
 	return m.Topic
 }
 
+// GetSubscriptionTopic returns the topic used for subscription fanout topic
+func (m *Model) GetSubscriptionTopic() string {
+	return topic.GetTopicName(m.ExtractedTopicProjectID, m.ExtractedSubscriptionName)
+}
+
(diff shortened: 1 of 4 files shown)


internal/subscription/model.go
```diff
@@ -81,6 +81,11 @@ func (m *Model) GetTopic() string {
 	return m.Topic
 }
 
+// GetSubscriptionTopic returns the topic used for subscription fanout topic
+func (m *Model) GetSubscriptionTopic() string {
+	return topic.GetTopicName(m.ExtractedTopicProjectID, m.ExtractedSubscriptionName)
```
razorao, line 86: Currently, the subscription topic is of form `/projects/topics/<subscription-name>`
Think of the scenario below:
I create a topic `/projects/topics/usecase-1234`
I create a subscription `/projects/topics/usecase-1234`

In this case, the subscription topic would be same as the ingestion topic?

It probably is a good idea to add some prefix/suffix to subscription-topic and ensure in validation that such topics can't be created directly by users.

razorao (approved):
LGTM apart from the naming of subscription topic

PR #280, opened 2021-12-06
Added Evaluation logic for Filter Feature

rutul25:
- Added Evaluation logic which will be called before delivering each message to check if the message satisfies the filter requirements if any

### DIFF
--- a/internal/subscriber/subscriber.go
+++ b/internal/subscriber/subscriber.go
@@ -571,8 +571,8 @@ func (s *Subscriber) pull(req *PullRequest) {
 				continue
 			}
 			ackID := ackMessage.BuildAckID()
-			// check if the subscription has any filter applied and check if the message satisfies it if any
 
+			// check if the subscription has any filter applied and check if the message satisfies it if any
 			if s.checkFilterCriteria(ctx, protoMsg) {
 				sm = append(sm, &metrov1.ReceivedMessage{AckId: ackID, Message: protoMsg, DeliveryAttempt: msg.CurrentRetryCount + 1})
 			} else {
@@ -592,6 +592,7 @@ func (s *Subscriber) pull(req *PullRequest) {
 	}()
 }
 
+// checks if the message satisfies filter criteria(if any) for the subscription
 func (s *Subscriber) checkFilterCriteria(ctx context.Context, protoMsg *metrov1.PubsubMessage) bool {
 	if s.subscription.FilterExpression != "" {
 		subFilter, err := s.subscription.GetFilterExpressionAsStruct()
--- a/internal/subscriber/subscriber.go
+++ b/internal/subscriber/subscriber.go
@@ -571,7 +571,37 @@ func (s *Subscriber) pull(req *PullRequest) {
 				continue
 			}
 			ackID := ackMessage.BuildAckID()
-			sm = append(sm, &metrov1.ReceivedMessage{AckId: ackID, Message: protoMsg, DeliveryAttempt: msg.CurrentRetryCount + 1})
+			// check if the subscription has any filter applied and check if the message satisfies it if any
+
+			isElligible := true
+
+			if s.subscription.FilterExpression != "" {
+				subFilter, err := s.subscription.GetFilterExpressionAsStruct()
+
+				if err != nil {
+					logger.Ctx(ctx).Errorw("subscriber: error in getting filter expression as a struct", "filter expression", s.subscription.FilterExpression,
+						"logfields", s.getLogFields(), "error", err.Error())
+				} else {
+					res, err := subFilter.Evaluate(protoMsg.Attributes)
+					if err != nil {
+						logger.Ctx(ctx).Errorw("subscriber: error occurred during filter evaluation", "filter expression", s.subscription.FilterExpression,
+							"logfields", s.getLogFields(), "error", err.Error())
+					} else {
+						if !res {
+							logger.Ctx(ctx).Infow("subscriber: Message didn't satisfy the filter criteria", "messageID", protoMsg.MessageId, "filter expression", s.subscription.FilterExpression,
+								"logfields", s.getLogFields())
+							isElligible = false
+						}
+					}
+				}
+			}
+
+			if isElligible {
+				sm = append(sm, &metrov1.ReceivedMessage{AckId: ackID, Message: protoMsg, DeliveryAttempt: msg.CurrentRetryCount + 1})
+			} else {
+				// self acknowledging the message as it does not need to be delivered for this subscription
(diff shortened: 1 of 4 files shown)


internal/subscriber/subscriber.go
```diff
@@ -571,7 +571,37 @@ func (s *Subscriber) pull(req *PullRequest) {
 				continue
 			}
 			ackID := ackMessage.BuildAckID()
-			sm = append(sm, &metrov1.ReceivedMessage{AckId: ackID, Message: protoMsg, DeliveryAttempt: msg.CurrentRetryCount + 1})
+			// check if the subscription has any filter applied and check if the message satisfies it if any
+
+			isElligible := true
```
sunnyagg, line 576: extract these to a separate method ?

internal/subscriber/subscriber_test.go
```diff
@@ -0,0 +1,203 @@
...
+		s.subscription.SetFilterExpression(test.FilterExpression)
+		request := &PullRequest{
+			ctx:              ctx,
+			MaxNumOfMessages: 10,
+		}
+		go s.pull(request)
+		select {
+		case data := <-respChan:
+			assert.Equal(t, len(data.ReceivedMessages), test.expectedNumOfMsgs)
+		case <-time.After(2 * time.Second):
```
sunnyagg, line 199: need better way to write unit tests, also how are the go routines getting terminated ?

internal/subscriber/subscriber.go
razorao: +1. Would suggest to write 1 method which accepts a message and returns bool if message should be delivered or not.

razorao: Also, the eligible/not eligible check is being done once the message is added to the offset management heap. It should be done before that.

internal/subscriber/subscriber.go
```diff
@@ -571,7 +571,37 @@ func (s *Subscriber) pull(req *PullRequest) {
...
+
+			if s.subscription.FilterExpression != "" {
+				subFilter, err := s.subscription.GetFilterExpressionAsStruct()
```
razorao, line 579: Sub filter can be extracted in the begining when the subscriber starts. Not in every `pull` call

internal/subscriber/subscriber.go
razorao: nit: isEligible

razorao (changes requested):
was assuming that we would just skip the messages that are not eligible for delivery. That way, we only commit messages that are eligible. Correct me if I'm wrong. Let's evaluate that approach as well.

internal/subscriber/subscriber.go
rutul25: To do that, we would then have to create a private field in subscriber to hold the filter_struct, which I think doesn't make sense as filter is a field of subscription and not subscriber or pass along the value in function as a parameter which isn't a better way either. 
Also this extraction is just an access call to a variable. It shouldn't be that costly operation. 

Do let me know if there is other way and added benefit of doing it the other way

rutul25:
> was assuming that we would just skip the messages that are not eligible for delivery. That way, we only commit messages that are eligible. Correct me if I'm wrong. Let's evaluate that approach as well.

The messages that doesn't satisfies the filter criteria needs to be committed for that subscriber right? To prevent it from getting it again from the broker. Let me know if I am missing something here.

internal/subscriber/subscriber.go
```diff
@@ -571,7 +571,14 @@ func (s *Subscriber) pull(req *PullRequest) {
...
 			}
 			ackID := ackMessage.BuildAckID()
-			sm = append(sm, &metrov1.ReceivedMessage{AckId: ackID, Message: protoMsg, DeliveryAttempt: msg.CurrentRetryCount + 1})
+
+			// check if the subscription has any filter applied and check if the message satisfies it if any
+			if s.checkFilterCriteria(ctx, protoMsg) {
+				sm = append(sm, &metrov1.ReceivedMessage{AckId: ackID, Message: protoMsg, DeliveryAttempt: msg.CurrentRetryCount + 1})
+			} else {
+				// self acknowledging the message as it does not need to be delivered for this subscription
+				s.acknowledge(ackMessage.(*AckMessage).WithContext(ctx))
```
sunnyagg, line 580: ack to subscriber means that message was consumed. shoudn't this offset be commited directly to broker ?

razorao:
> > was assuming that we would just skip the messages that are not eligible for delivery. That way, we only commit messages that are eligible. Correct me if I'm wrong. Let's evaluate that approach as well.
> 
> The messages that doesn't satisfies the filter criteria needs to be committed for that subscriber right? To prevent it from getting it again from the broker. Let me know if I am missing something here.

@rutul25  Was thinking in lines of what @sunnyagg has commented. Was assuming the message won't reach the subscriber at all.

internal/subscriber/subscriber.go
rutul25: So to do that, we will have to handle the whole filtering in the consumer itself right? And if so, we will have to handle filtering logic for all the consumer implementations in future. It will lead to having same thing handled at multiple places. That's why I handled the filtering in subscriber, after the consumer has fetched the messages from the broker. 

Thoughts on this?

rutul25: Or we can handle the filtering in subscriber only, but before the offset management happen in subscriber and commit the skipped messages directly to consumer from there.

rutul25:
> > > was assuming that we would just skip the messages that are not eligible for delivery. That way, we only commit messages that are eligible. Correct me if I'm wrong. Let's evaluate that approach as well.
> > 
> > 
> > The messages that doesn't satisfies the filter criteria needs to be committed for that subscriber right? To prevent it from getting it again from the broker. Let me know if I am missing something here.
> 
> @rutul25 Was thinking in lines of what @sunnyagg has commented. Was assuming the message won't reach the subscriber at all.

Had a discussion with @sunnyagg  regarding this. Decided to go with the current approach because of this reason

- Let there be 3 messages - Msg-1, Msg-2 and Msg-3. Msg-1 and Msg-3 satisfies the filter criteria while Msg-3 doesn't.
- If we directly ack Msg-2 using brokerStore consumer, in Kafka, new committed offset will be set to 2.
- Now if for some reason, delivery of Msg-1 fails, we will not be able to retry it as the broker's committed offset is already set to 2.
- To avoid this, subscriber Acknowledge is used which will wait for Msg-1 status before committing the offsets.

sunnyagg (approved):

vnktram (approved):
LGTM

razorao (approved):

PR #284, opened 2021-12-13
Consume/deployment

vnktram:
- [x] Fetch ordinal ID from host name
- [x] Load ordinal ID to consumeConfig. Part of #283

### DIFF
--- a/build/docker/entrypoint.sh
+++ b/build/docker/entrypoint.sh
@@ -10,6 +10,8 @@ initialize()
       return
     fi
 
+    export METRO_CONSUMEPLANE_ORDINAL_ID=${HOSTNAME##*-}
+
     mkdir -p /app/configs
     touch /app/configs/ca-cert.pem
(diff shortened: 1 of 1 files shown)


build/docker/entrypoint.sh
```diff
@@ -10,6 +10,8 @@ initialize()
       return
     fi
 
+    export METRO_CONSUMEPLANE_ORDINALID=${HOSTNAME##*-}
```
razorao, line 13: Did not understand the `##*-`.
Please explain.

vnktram: https://github.com/kubernetes/kubernetes/issues/40651
We need to fetch the ordinal ID for each pod of the stateful set.
Currently, k8s downstream APIs do not expose this field.
Hence the workaround - https://github.com/kubernetes/kubernetes/issues/40651#issuecomment-623593113

PR #285, opened 2021-12-14
#256: Ordering documentation

razorao:


docs/docs/guides/receiving/ordering.md
```diff
@@ -0,0 +1,64 @@
...
+        {
+            "data": "YQ==",
+            "orderingKey": "o1"
+        }
+    ]
+}
+```
+Messages having the same ordering key are delivered in the order Metro receives them. The order of messages delivered across multiple ordering keys can be different from the order in which they are published.
+
+## Publisher Syncronization
```
sunnyagg, line 54: probably recommend adding `outbox pattern` as solution for publishing in correct order

sunnyagg (approved):

PR #289, opened 2021-12-17
Fixed the retry commit bug

razorao:
Current behaviour of retry function:
- Mod ack request is received
- Retry is attempted
- The message is commited to the source topic

New Behaviour:
- Mod Ack request is received
- Retry is attempted
- Retry calls the `commitAndRemoveFromMemory`
- Ack method also calls the same method on receiving an Ack request

### DIFF
--- a/internal/subscriber/subscriber.go
+++ b/internal/subscriber/subscriber.go
@@ -110,18 +110,90 @@ func (s *Subscriber) retry(ctx context.Context, msg messagebroker.ReceivedMessag
 		return
 	}
 
-	// commit on the primary topic after message has been submitted for retry
-	_, err = s.consumer.CommitByPartitionAndOffset(ctx, messagebroker.CommitOnTopicRequest{
-		Topic:     msg.Topic,
-		Partition: msg.Partition,
-		// add 1 to current offset
-		// https://docs.confluent.io/5.5.0/clients/confluent-kafka-go/index.html#pkg-overview
-		Offset: msg.Offset + 1,
-	})
-	if err != nil {
-		logger.Ctx(ctx).Errorw("subscriber: failed to commit message", "logFields", s.getLogFields(), "error", err.Error())
-		s.errChan <- err
-		return
+	s.commitAndRemoveFromMemory(ctx, msg)
+}
+
+func (s *Subscriber) commitAndRemoveFromMemory(ctx context.Context, msg messagebroker.ReceivedMessage) {
+	logFields := s.getLogFields()
+
+	tp := TopicPartition{topic: msg.Topic, partition: msg.Partition}
+	stats := s.consumedMessageStats[tp]
+
+	offsetToCommit := msg.Offset
+	shouldCommit := false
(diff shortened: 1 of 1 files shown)


vnktram (approved):
LGTM. Can we capture how this flow would change once we allow modack requests to carry a non-zero deadline as well?

razorao:
> LGTM. Can we capture how this flow would change once we allow modack requests to carry a non-zero deadline as well?

There would be no change. It's already handled in the current code. All we need to do in case of non zero acks is update the heap with new deadline. This is already being done.

internal/subscriber/subscriber.go
```diff
@@ -249,12 +189,68 @@ func (s *Subscriber) acknowledge(req *AckMessage) {
...
+	// check if message is present in-memory or not
+	if _, ok := stats.consumedMessages[msgID]; !ok {
+		logger.Ctx(ctx).Infow("subscriber: skipping ack as message not found in-memory", "logFields", logFields)
+		return
+	}
+
+	msg := stats.consumedMessages[msgID].(messagebroker.ReceivedMessage)
+
+	// if somehow an ack request comes for a message that has met deadline eviction threshold
+	if req.HasHitDeadline() {
```
sunnyagg, line 243: shouldn't this validation check happen outside of subscriber acknowledge, probably at controller layer ?

even if we need the check, i guess we should throw error here and let the caller handle it whether to retry or not

internal/subscriber/subscriber.go
```diff
@@ -110,80 +110,20 @@ func (s *Subscriber) retry(ctx context.Context, msg messagebroker.ReceivedMessag
...
-		// add 1 to current offset
-		// https://docs.confluent.io/5.5.0/clients/confluent-kafka-go/index.html#pkg-overview
-		Offset: msg.Offset + 1,
-	})
-	if err != nil {
-		logger.Ctx(ctx).Errorw("subscriber: failed to commit message", "logFields", s.getLogFields(), "error", err.Error())
-		s.errChan <- err
-		return
-	}
+	s.commitAndRemoveFromMemory(ctx, msg)
```
sunnyagg, line 113: can we call acknowledge directly here ?

razorao: Acknowledge:  

    1. Do some checks
    2. Commit continous offsets
    3. Remove from memory
    4. Update the metrics related to acks

Have extracted points 2 and 3 (the common part across ack, deadline, nack)

internal/subscriber/subscriber.go
razorao: So this check exists to check if a message has exceeded deadline and not been evicted by our ticker. In this case we throw it for retry. 

Currently the subscriber is the only entity which makes decisions on messages. I guess that is the reason why this is not modelled as you've said. 

Also we don't throw errors in case of acks and nacks today. Perhaps the reason is pubsub does not do it that way in case of streaming requests?

sunnyagg (approved):

PR #295, opened 2022-01-06
Added List APIs for topic and subscriptions

rutul25:
Added following APIs

- List all topics for the given project.
- List all subscriptions for the given project.
- List all subscriptions created on the given topic.

### DIFF
--- a/service/web/publisherserver.go
+++ b/service/web/publisherserver.go
@@ -93,6 +93,30 @@ func (s publisherServer) DeleteTopic(ctx context.Context, req *metrov1.DeleteTop
 	return &emptypb.Empty{}, nil
 }
 
+func (s publisherServer) ListProjectTopics(ctx context.Context,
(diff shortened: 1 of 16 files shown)


service/web/publisherserver.go
```diff
@@ -93,6 +93,30 @@ func (s publisherServer) DeleteTopic(ctx context.Context, req *metrov1.DeleteTop
...
+
+	topics, err := s.topicCore.List(ctx, topic.Prefix+req.ProjectId)
+
+	if err != nil {
+		return nil, merror.ToGRPCError(err)
+	}
+
+	res := []string{}
+	for _, t := range topics {
+		if !t.IsDeadLetterTopic() {
```
rutul25, line 113: Here, I have currently filtered out DLQ topics from the result. Trying to figure out a good way to do the same for delay topics. Will append changes for the same in next revision

vnktram (approved):
LGTM

PR #297, opened 2022-01-10
Add openapi server handler to web deployment

vnktram:
- [x] Migrate openapi server to be served as part of web deployment


rutul25 (approved):
LGTM

PR #305, opened 2022-02-07
Added more metrics on subscription

rutul25:
Following metrics are added

- Number of retained acked messages
- Total size in bytes of retained acked messages
- Number of messages pushed in dlq

### DIFF
--- a/internal/subscriber/metrics.go
+++ b/internal/subscriber/metrics.go
@@ -26,6 +26,8 @@ var (
 	subscriberTimeTakenToRemoveMsgFromMemory   *prometheus.HistogramVec
 	subscriberTimeTakenToIdentifyNextOffset    *prometheus.HistogramVec
 	subscriberTimeTakenToPushToRetry           *prometheus.HistogramVec
+	subscriberNumberOfReatinedAckedMessages    *prometheus.GaugeVec
+	subscriberReatinedAckedMessagesSize        *prometheus.GaugeVec
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


internal/subscriber/retry/metrics.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	env                                string
+	subscriberTotalMessagesPushedToDLQ *prometheus.GaugeVec
+)
+
+func init() {
+	env = os.Getenv("APP_ENV")
```
vnktram, line 16: We have to refactor this part. Fetching Env from the underlying machine is an expensive call and we repeat this over every metric component across all packages. We should look at passing this down from the driver instead.

vnktram: Probably a refactor we should address in a separate task.

internal/subscriber/subscriberimpl.go
```diff
@@ -182,6 +182,8 @@ func (s *BasicImplementation) Acknowledge(ctx context.Context, req *AckMessage,
 
 	subscriberMessagesAckd.WithLabelValues(env, s.topic, s.subscription.Name, s.subscriberID).Inc()
 	subscriberTimeTakenToAckMsg.WithLabelValues(env, s.topic, s.subscription.Name).Observe(time.Now().Sub(msg.PublishTime).Seconds())
+	subscriberNumberOfReatinedAckedMessages.WithLabelValues(env, s.topic, s.subscription.Name).Inc()
```
vnktram, line 185: nit: typo Reatined->Retained

internal/subscriber/metrics.go
```diff
@@ -26,6 +26,8 @@ var (
 	subscriberTimeTakenToRemoveMsgFromMemory   *prometheus.HistogramVec
 	subscriberTimeTakenToIdentifyNextOffset    *prometheus.HistogramVec
 	subscriberTimeTakenToPushToRetry           *prometheus.HistogramVec
+	subscriberNumberOfReatinedAckedMessages    *prometheus.GaugeVec
```
vnktram, line 29: nit: typo Reatined->Retained

internal/subscriber/metrics.go
```diff
@@ -26,6 +26,8 @@ var (
...
+	subscriberReatinedAckedMessagesSize        *prometheus.GaugeVec
```
vnktram, line 30: typo^

vnktram (changes requested):

vnktram (approved):
LGTM

PR #321, opened 2022-03-23
Move prod cache driver to redis

vnktram:


config/prod.toml
```diff
@@ -61,7 +61,7 @@
             caPem              = ""
 
 [cache]
-    driver                    = "consul"
+    driver                    = "redis"
```
sunnyagg, line 64: shouldn't you also define the redisconfig with the drive

vnktram: This is being defined via kubestash. Please refer [https://razorpay.atlassian.net/browse/BAD-7487](https://razorpay.atlassian.net/browse/BAD-7487)
The base config is in default.yaml

sunnyagg (approved):

PR #336, opened 2022-05-12
Added Project core test cases: Increased coverage to 80%

akashgoyal96:


internal/project/core_test.go
```diff
@@ -53,3 +56,121 @@ func TestCore_ExistsWithID(t *testing.T) {
...
+			c := &Core{
+				repo: tt.fields.repo,
+			}
+			var err2 error = nil
+			if len(tt.args.projectID) == 0 {
+				err2 = fmt.Errorf("Invalid Project ID!")
+			}
+			mod := &Model{}
+			mockRepo.EXPECT().Get(gomock.Any(), gomock.Any(), mod).Do(func(arg1 context.Context, arg2 string, mod *Model) {
+				mod.SetVersion("123")
```
vnktram, line 119: In case of an invalid project ID, this will still set a version value for the Model. Please fix this.

internal/project/core_test.go
```diff
@@ -53,3 +56,121 @@ func TestCore_ExistsWithID(t *testing.T) {
...
+			},
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &Core{
+				repo: tt.fields.repo,
+			}
+			mockRepo.EXPECT().Exists(gomock.Any(), project.Key()).Return(true, nil)
```
vnktram, line 169: Please test the flow where the project does not exist.

vnktram (changes requested):
Please address comments

internal/project/core_test.go
```diff
@@ -53,3 +57,198 @@ func TestCore_ExistsWithID(t *testing.T) {
...
+			},
+			args: args{
+				ctx:       ctx,
+				projectID: project2.ProjectID,
+			},
+			want:    project2,
+			wantErr: false,
+		},
+		{
+			name: "Get Project 3 Successfully.",
```
vnktram, line 124: Why are we testing the same case thrice? Can we randomly pick a number between 1 and 3 and see if it picked the correct project ID?

internal/project/core_test.go
```diff
@@ -53,3 +57,198 @@ func TestCore_ExistsWithID(t *testing.T) {
...
+			c := &Core{
+				repo: tt.fields.repo,
+			}
+			var err2 error = nil
+			if len(tt.args.projectID) == 0 {
+				err2 = fmt.Errorf("Invalid Project ID!")
+			}
+			mockRepo.EXPECT().Get(gomock.Any(), gomock.Any(), &Model{}).Do(func(arg1 context.Context, arg2 string, mod *Model) {
+				if err2 == nil {
+					if arg2 == common.GetBasePrefix()+Prefix+"testID1" {
```
vnktram, line 159: tests need to be generic. Please avoid repetitions.

vnktram (changes requested):

internal/project/core_test.go
```diff
@@ -53,3 +58,177 @@ func TestCore_ExistsWithID(t *testing.T) {
...
+				ctx:       ctx,
+				projectID: "",
+			},
+			want:    nil,
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &Core{
```
sakshi-goyal-razorpay, line 139: this initialisation is same for all tests.. why are we initialising it again..?

akashgoyal96: These are auto-generated from the command I use for generating tests for individual functions. I am only constructing the test cases and the mocks.

internal/project/core_test.go
```diff
@@ -53,3 +58,177 @@ func TestCore_ExistsWithID(t *testing.T) {
...
+		Labels:    map[string]string{"label": "value"},
+	}
+	project1.SetVersion("1")
+	project2.SetVersion("2")
+	project3.SetVersion("3")
+
+	num := rand.Intn(3)
+
+	testProject := &Model{}
+	switch num {
```
vnktram, line 96: Use a map instead of switch for deciding which project to choose from.

vnktram, line 96: Use a map instead of switch for deciding which protect to choose from.

internal/project/core_test.go
```diff
@@ -53,3 +58,177 @@ func TestCore_ExistsWithID(t *testing.T) {
...
+			args: args{
+				ctx: ctx,
+				m:   dummyModel,
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &Core{
```
vnktram, line 218: Move initialization outside the loop. Core is common for all iterations

vnktram, line 218: Move initialization outside the loop. Core is common for all iterations

internal/project/core_test.go
```diff
@@ -53,3 +58,177 @@ func TestCore_ExistsWithID(t *testing.T) {
...
+				repo: tt.fields.repo,
+			}
+			var err2 error = nil
+			if len(tt.args.m.ProjectID) == 0 {
+				err2 = fmt.Errorf("Invalid Project ID!")
+				mockRepo.EXPECT().Exists(gomock.Any(), tt.args.m.Key()).Return(false, err2)
+			} else {
+				mockRepo.EXPECT().Exists(gomock.Any(), tt.args.m.Key()).Return(true, err2)
+				mockRepo.EXPECT().Delete(gomock.Any(), tt.args.m).Return(err2)
```
vnktram, line 227: This should return `nil` and not err2. err2 is contextual to `Exists`

vnktram, line 227: This should return `nil` and not err2. err2 is contextual to `Exists`

vnktram (changes requested):

internal/project/core_test.go
akashgoyal96: But if I have initialized err2 = nil irrespective, then shouldn't this work.

vnktram (approved):

PR #339, opened 2022-05-18
Store subscriber/consumer message processing timestamps

sakshi-goyal-razorpay:


internal/subscriber/subscriber.go
```diff
@@ -98,6 +99,11 @@ func (s *Subscriber) GetModAckChannel() chan *ModAckMessage {
 	return s.modAckChan
 }
 
+// UpdateLastProcessingTime updates the last processing time of the subscriber
+func (s *Subscriber) UpdateLastProcessingTime() {
```
vnktram, line 103: Can we not expose this as an exported function but instead call it from both ack and nack methods within subscriber?

vnktram (approved):
LGTM. Added a suggestion, address it at your discretion

PR #342, opened 2022-05-19
Improve Web Credentials + Subscription Test coverage

akashgoyal96:


internal/credentials/repo_test.go
```diff
@@ -0,0 +1,81 @@
...
+				ctx:    ctx,
+				prefix: "sample",
+			},
+			want:    topics,
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := &Repo{
```
vnktram, line 63: Move initializations outside the loop.

internal/credentials/model_test.go
```diff
@@ -45,3 +46,41 @@ func TestModel_HiddenPassword_Failure(t *testing.T) {
 	assert.Empty(t, hiddenPwd)
 	assert.Equal(t, err, ErrPasswordNotInExpectedFormat)
 }
+
+// func TestNewCredential(t *testing.T) {
```
vnktram, line 50: Can we avoid adding commented code as part of the PR?

internal/credentials/repo_test.go
```diff
@@ -0,0 +1,81 @@
...
+			name: "Get Models list with version set",
+			fields: fields{
+				BaseRepo: mockRepo,
+			},
+			args: args{
+				ctx:    ctx,
+				prefix: "sample",
+			},
+			want:    topics,
+			wantErr: false,
```
vnktram, line 58: Can we add another test here to verify the error case as well?

vnktram (changes requested):

PR #344, opened 2022-05-24
Introduce managed pool for workers

vnktram:


internal/stream/stream.go
```diff
@@ -103,18 +116,20 @@ func (ps *PushStream) processMessages() {
...
-	pullBatchSize := 10
 	if ps.subscription.EnableMessageOrdering {
 		pullBatchSize = 1
 	}
 
 	// Send message pull request to subsriber request channel
 	// logger.Ctx(ctx).Debugw("worker: sending a subscriber pull request", "logFields", ps.getLogFields())
+	if ps.counter >= int64(pullBatchSize) {
+		return
+	}
```
sakshi-goyal-razorpay, line 127: why do we need to explicitly stop adding msgs to msg channel..? isn't ants pool will take care of only spawning go routines limited to max pool size..?

sakshi-goyal-razorpay (approved):
LGTM, just need clarification on one small thing.

internal/stream/stream.go
vnktram: The number of inflight messages is something we want to actively manage. If we do not regulate the count then we end up consuming constantly from kafka and the rate at which we deliver will always be lower than the rate at which we consume from Kafka.

vnktram: If the number of outstanding messages is less than the number of goroutines available in the pool then we fetch, else we wait for the goroutines to be freed up.

PR #345, opened 2022-05-26
subscriber level healthchecks

sakshi-goyal-razorpay:
1. Adding delay consumer errors to subscriber err channel
2. Restart subscriber for any error on its err channel
3. Restart stream for any error while running stream

### DIFF
--- a/internal/stream/metrics.go
+++ b/internal/stream/metrics.go
@@ -15,7 +15,7 @@ var (
 	workerPushEndpointTimeTaken      *prometheus.HistogramVec
 	workerSubscriberErrors           *prometheus.CounterVec
 	workerPushEndpointCallsCount     *prometheus.CounterVec
-	workerEntityRestartCount         *prometheus.CounterVec
+	workerComponentRestartCount      *prometheus.CounterVec
 )
 
 func init() {
@@ -47,7 +47,7 @@ func init() {
 		Name: "metro_worker_push_endpoint_http_calls",
 	}, []string{"env", "topic", "subscription", "endpoint", "subscriberId"})
 
-	workerEntityRestartCount = promauto.NewCounterVec(prometheus.CounterOpts{
-		Name: "metro_worker_subscriber_restart_count",
-	}, []string{"env", "entity", "topic", "subscription"})
+	workerComponentRestartCount = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "metro_worker_component_restart_count",
+	}, []string{"env", "component", "topic", "subscription"})
 }
--- a/internal/stream/metrics.go
+++ b/internal/stream/metrics.go
@@ -15,6 +15,7 @@ var (
 	workerPushEndpointTimeTaken      *prometheus.HistogramVec
 	workerSubscriberErrors           *prometheus.CounterVec
 	workerPushEndpointCallsCount     *prometheus.CounterVec
+	workerEntityRestartCount         *prometheus.CounterVec
 )
 
 func init() {
@@ -45,4 +46,8 @@ func init() {
 	workerPushEndpointCallsCount = promauto.NewCounterVec(prometheus.CounterOpts{
 		Name: "metro_worker_push_endpoint_http_calls",
 	}, []string{"env", "topic", "subscription", "endpoint", "subscriberId"})
+
+	workerEntityRestartCount = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "metro_worker_subscriber_restart_count",
+	}, []string{"env", "entity", "topic", "subscription"})
 }
--- a/internal/stream/stream.go
+++ b/internal/stream/stream.go
@@ -95,9 +95,8 @@ func (ps *PushStream) Start() error {
 			case err = <-ps.subs.GetErrorChannel():
 				// if channel is closed, this can return with a nil error value
 				if err != nil {
-					logger.Ctx(ps.ctx).Errorw("worker: error from subscriber", "logFields", ps.getLogFields(), "error", err.Error())
+					logger.Ctx(ps.ctx).Errorw("worker: error from subscriber, restarting", "logFields", ps.getLogFields(), "error", err.Error())
 					workerSubscriberErrors.WithLabelValues(env, ps.subscription.ExtractedTopicName, ps.subscription.Name, err.Error(), ps.subs.GetID()).Inc()
-					logger.Ctx(ps.ctx).Infow("worker: restarting subscriber", "logFields", ps.getLogFields())
 					if err = ps.restartSubsciber(); err != nil {
 						ps.restartChan <- true
 						return err
@@ -173,7 +172,7 @@ func (ps *PushStream) Restart(ctx context.Context) {
 		logger.Ctx(ctx).Errorw(
 			"worker: push stream stop error",
 			"subscription", ps.subscription.Name,
-			"error", err,
+			"error", err.Error(),
 		)
 		return
 	}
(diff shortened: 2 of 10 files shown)


internal/stream/metrics.go
```diff
@@ -45,4 +46,8 @@ func init() {
 	workerPushEndpointCallsCount = promauto.NewCounterVec(prometheus.CounterOpts{
 		Name: "metro_worker_push_endpoint_http_calls",
 	}, []string{"env", "topic", "subscription", "endpoint", "subscriberId"})
+
+	workerEntityRestartCount = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "metro_worker_subscriber_restart_count",
+	}, []string{"env", "entity", "topic", "subscription"})
```
vnktram, line 52: suggestion: rename entity to component

internal/subscriber/core.go
```diff
@@ -51,6 +54,8 @@ func (c *Core) NewSubscriber(ctx context.Context,
 	}
 
 	subsCtx, cancelFunc := context.WithCancel(ctx)
+	errChan := make(chan error, DefaultLimitForErrChannel)
```
vnktram, line 57: Why do we want to use a buffered channel here? Once an error is encountered we restart the subscriber anyway.

sakshi-goyal-razorpay: existing code was creating a buffered channel, here I have just passed it to delay consumer.
But yeah, since we are reading from it at the same time in stream, we can use a unbuffered channel

internal/tasks/subscriptiontask.go
```diff
@@ -236,6 +236,19 @@ func (sm *SubscriptionTask) handleNodeBindingUpdates(ctx context.Context, newBin
...
 						"error", err.Error())
 				}
+
+				for {
+					select {
+					case <-ctx.Done():
+						return
+					case restart := <-handler.GetRestartChannel():
+						if restart {
+							handler.Restart(ctx)
```
vnktram, line 246: Why send a false and true and check here? Send a signal on restartChan only if a restart is required.

vnktram (changes requested):

internal/stream/stream.go
```diff
@@ -78,12 +85,19 @@ func (ps *PushStream) Start() error {
 				// stop the subscriber after all the send channels are closed
 				ps.stopSubscriber()
 
+				ps.restartChan <- false
```
vnktram, line 88: Can we avoid this?

vnktram (changes requested):

internal/tasks/subscriptiontask.go
sakshi-goyal-razorpay: if a push stream was running but due to node binding change, we stop it, then we don't need to restart it. 
In that case, only push stream context is cancelled, and not the subscription task's.
To differentiate b/w these, had to send false on restart Chan to exit the start go routine

internal/stream/stream.go
```diff
@@ -84,6 +97,11 @@ func (ps *PushStream) Start() error {
 				if err != nil {
 					logger.Ctx(ps.ctx).Errorw("worker: error from subscriber", "logFields", ps.getLogFields(), "error", err.Error())
```
vnktram, line 98: Consecutive log lines. Can we merge them into one?

internal/stream/stream.go
```diff
@@ -94,6 +112,19 @@ func (ps *PushStream) Start() error {
 	return errGrp.Wait()
 }
 
+func (ps *PushStream) restartSubsciber() error {
+	ps.subs.Stop()
+	var err error
+	ps.subs, err = ps.subscriberCore.NewSubscriber(context.Background(), ps.nodeID, ps.subscription, defaultTimeoutMs,
```
vnktram, line 118: How is start handled here?

internal/stream/stream.go
```diff
@@ -134,6 +165,31 @@ func (ps *PushStream) Stop() error {
...
 
+// Restart is used to restart the push subscription processing
+func (ps *PushStream) Restart(ctx context.Context) {
+	logger.Ctx(ps.ctx).Infow("worker: push stream restart invoked", "subscription", ps.subscription.Name)
+	err := ps.Stop()
+	if err != nil {
+		logger.Ctx(ctx).Errorw(
+			"worker: push stream stop error",
+			"subscription", ps.subscription.Name,
+			"error", err,
```
vnktram, line 176: Print err.Error() instead

internal/stream/stream.go
```diff
@@ -134,6 +165,31 @@ func (ps *PushStream) Stop() error {
...
+		)
+		return
+	}
+	go func(ctx context.Context) {
```
vnktram, line 180: Do not run goroutines that do not exit on ctx.Done() and other signals. these have the potential to end up being zombie routines.

internal/stream/stream.go
```diff
@@ -358,6 +414,8 @@ func NewPushStream(ctx context.Context, nodeID string, subName string, subscript
 		subscriberCore:   subscriberCore,
 		doneCh:           make(chan struct{}),
 		httpClient:       httpclient,
+		restartChan:      make(chan bool, 10),
```
vnktram, line 417: Both channels are signal-based channels. Avoid buffered channels for this use case.

internal/subscriber/retry/delayconsumer.go
```diff
@@ -80,6 +82,7 @@ func (dc *DelayConsumer) Run(ctx context.Context) {
 			if err != nil {
 				if !messagebroker.IsErrorRecoverable(err) {
 					logger.Ctx(dc.ctx).Errorw("delay-consumer: error in receiving messages", dc.LogFields("error", err.Error())...)
+					dc.errChan <- err
```
vnktram, line 85: This PR aims to restart subscribers on any and every unrecoverable error. Ideally, we want to restart the failed component if possible in the future. Keep this in mind when you plan for the subscriber re-arch

internal/tasks/subscriptiontask.go
```diff
@@ -194,10 +194,8 @@ func (sm *SubscriptionTask) handleNodeBindingUpdates(ctx context.Context, newBin
 			if handler, ok := sm.pushHandlers.Load(oldKey); ok && handler != nil {
 				logger.Ctx(ctx).Infow("handler found, calling stop", "key", oldKey)
 				go func(ctx context.Context) {
```
vnktram, line 196: Can we avoid invoking a goroutine here?

vnktram (changes requested):

internal/stream/stream.go
sakshi-goyal-razorpay: inside new Subscriber, core module is running it. https://github.com/razorpay/metro/blob/master/internal/subscriber/core.go#L126

internal/tasks/subscriptiontask.go
sakshi-goyal-razorpay: handler.Start is a waiting call..
https://github.com/razorpay/metro/blob/master/internal/stream/stream.go#L94
It will not exit until the stream exits.

internal/stream/stream.go
sakshi-goyal-razorpay: Start is a waiting call.. If ctx is done, then start will exit, and so will this go routine

internal/stream/stream.go
sakshi-goyal-razorpay: Have make these unbuffered, but in case ctx is done, I am calling stop() directly rather than sending signal on stop chan, as the stream manager also exists when parent ctx is done

internal/stream/stream.go
sakshi-goyal-razorpay: The go routine just runs the Start of subscriber, and only when the ctx is done, the subscriber go routine is exited, otherwise it will keep on listening on various channels.

vnktram (approved):
We're unsure of how this would behave in the long run. Let's test this on stage for a day or two before we push this to prod

PR #349, opened 2022-05-31
 Reset offset to start of topic for dlq subscriptions

sakshi-goyal-razorpay:


internal/subscriber/consumer.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/razorpay/metro/internal/brokerstore"
+	tp "github.com/razorpay/metro/internal/topic"
```
vnktram, line 8: Why not use `topic` instead of tp?

internal/subscriber/consumer.go
```diff
@@ -47,12 +48,17 @@ type consumerManager struct {
 // NewConsumerManager ...
 func NewConsumerManager(ctx context.Context, bs brokerstore.IBrokerStore, brokerTimeout int, subscriberID string, subscriptionName,
 	topicName, retryTopicName string) (IConsumer, error) {
+	autoOffsetReset := "latest"
+	if tp.IsDLQTopic(topicName) {
+		autoOffsetReset = "earliest"
```
vnktram, line 53: Every time a new consumer is instantiated for a DLQ topic, it will retry messages from the start of the topic. Ideally, the offset must be set once during the time of DLQ subscription creation.

internal/subscriber/consumer.go
vnktram: Please rename the variable there. Package names take precedence over variables

internal/subscriber/consumer.go
sakshi-goyal-razorpay: auto offset reset is only used if there is no committed offset. Even if consumer dies, and next consumer will start from last committed offset. In case there is no committed offset, means it is the first time we are consuming from the topic for that consumer group, then only this config will be used.
https://stackoverflow.com/questions/32390265/what-determines-kafka-consumer-offset

vnktram (approved):

PR #351, opened 2022-06-01
Fix missing topic panic

vnktram:


internal/tasks/schedulertask.go
```diff
@@ -361,6 +361,7 @@ func (sm *SchedulerTask) refreshNodeBindings(ctx context.Context) error {
 		validBindings[subPart] = nb
 	}
 
+	logger.Ctx(ctx).Infow("schedulertask: fetched valid bindings after removing invalid bindings", "validBindings", validBindings)
```
sakshi-goyal-razorpay, line 364: we have already merged the missing topic issue right..? This is just logging valid bindings.. PR title seems to be confusing.

PR #360, opened 2022-06-14
Metro 43 update sub topic partitions

akashgoyal96:


service/web/adminserver.go
```diff
@@ -121,6 +129,19 @@ func (s adminServer) ModifyTopic(ctx context.Context, req *metrov1.AdminTopic) (
 		return nil, merror.ToGRPCError(uerr)
 	}
 
+	// Fetch Updated Topic
+	updatedTopic, err := s.topicCore.Get(ctx, m.Name)
+	if err != nil {
+		return nil, merror.ToGRPCError(err)
+	}
```
sakshi-goyal-razorpay, line 136: why do we need to get this again..?
we can move  s.subscriptionCore.RescaleSubTopics(ctx, updatedTopic) call inside 
if int(req.NumPartitions) != existingTopic.NumPartitions block only, right..?

sakshi-goyal-razorpay (commented):
Two things:
1. Lets add a functional test case to check if rescaling actually works.
2. Why have we updated submodule commit.?

internal/project/model.go
```diff
@@ -26,3 +29,14 @@ func (m *Model) Key() string {
 func (m *Model) Prefix() string {
 	return common.GetBasePrefix() + Prefix
 }
+
+// FetchProjectID returns the key for storing in the registry
+func FetchProjectID(val string) string {
+	stringArr := strings.Split(val, "/")
+	arrLen := len(stringArr)
+	if arrLen < 1 {
+		log.Fatalf("Invalid ProjectID given as input: [%v]", val)
```
vnktram, line 38: Please use the logger package from `pkg` folder. Do not use `log` directly.

internal/subscription/core.go
```diff
@@ -344,6 +348,106 @@ func createDelayTopics(ctx context.Context, m *Model, topicCore topic.ICore, top
...
+		}
+		completeSubList = append(completeSubList, subList...)
+	}
+	admin, err := c.brokerStore.GetAdmin(ctx, messagebroker.AdminClientOptions{})
+	if err != nil {
+		return merror.ToGRPCError(err)
+	}
+
+	for _, m := range completeSubList {
+		if m.ExtractedTopicName != topicModel.ExtractedTopicName {
```
vnktram, line 372: The comparison should be between `sub.Topic` and `topicModel.Name`. Extracted names ignore project for the topic

internal/project/model.go
akashgoyal96: Sure, updating this. We should change this similarly in different places where this is done which I took as reference.

service/web/adminserver.go
akashgoyal96: Still we would need to fetch the updated topic right to pass it into the RescaleSubTopic arguments.

sakshi-goyal-razorpay:
> Two things:
> 
> 1. Lets add a functional test case to check if rescaling actually works.
> 2. Why have we updated submodule commit.?

what about these..?

PR #371, opened 2022-06-23
Implement graceful shutdown

sakshi-goyal-razorpay:


cmd/service/metro/metro.go
```diff
@@ -104,6 +105,7 @@ func Run(ctx context.Context) {
 	go func() {
 		sig := <-sigCh
 		logger.Ctx(ctx).Infow("received a signal, stopping metro", "signal", sig)
+		component.GracefulShutdown(fmt.Errorf("OS Signal error"))
```
vnktram, line 108: Why not send a signal to the error channel from here? Is the function intended to achieve any non-blocking flows?

internal/brokerstore/brokerstore.go
```diff
@@ -294,3 +299,13 @@ func findAllMatchingKeyPrefix(mp *sync.Map, prefix string) []interface{} {
 	})
 	return values
 }
+
+// FlushAllProducers will iterate over the producer map, and flush all messages in the producer buffer
+func (b *BrokerStore) FlushAllProducers() bool {
+	b.producerMap.Range(func(key, producer interface{}) bool {
+		producer.(messagebroker.Producer).Flush(defaultFlushTimeoutMs)
```
vnktram, line 306: This is a sequential call. Which would add up if the num of producers are high. Please make it parallel.

pkg/messagebroker/kafka.go
```diff
@@ -762,3 +762,8 @@ func (k *KafkaBroker) Shutdown(ctx context.Context) {
 func (k *KafkaBroker) IsClosed(_ context.Context) bool {
 	return k.isProducerClosed
 }
+
+// Flush flushes the producer buffer
+func (k *KafkaBroker) Flush(timeoutMs int) {
+	k.Producer.Flush(timeoutMs)
```
vnktram, line 768: FLush is a function of the producer client and not consumer/admin clients. Let's restrict flush operations only to producers

sakshi-goyal-razorpay: All I can do is add a producer null check over here..Is that what you mean..?

cmd/service/metro/metro.go
sakshi-goyal-razorpay: I created a new method, in case we want to add something else to gracefully shutdown a component, then we can add that in the same method, rather than having all those lines added in this go routine.

akashgoyal96 (approved):

vnktram (approved):

PR #385, opened 2022-07-05
GetSubscription API added to subscriptionserver

sumitkrzp:


service/web/subscriberserver_test.go
```diff
@@ -684,3 +685,85 @@ func TestSubscriberServer_ListProjectSubscriptionsFailure(t *testing.T) {
...
+		RetryPolicy: &subscription.RetryPolicy{
+			MinimumBackoff: 20,
+			MaximumBackoff: 300,
+		},
+	}
+	subscriptionCore.EXPECT().Get(gomock.Any(), req.GetName()).Times(1).Return(sub, nil)
+	res, err := server.GetSubscription(ctx, req)
+	assert.NotNil(t, res)
+	assert.Nil(t, err)
+	assert.Equal(t, sub.Name, res.Name)
```
sakshi-goyal-razorpay, line 730: can we use reflection for comparison..?

service/web/server_test.go
```diff
@@ -103,3 +104,32 @@ func TestGetProjectIDFromRequest(t *testing.T) {
...
+		request interface{}
+		name    string
+		err     error
+	}
+	type testReq struct {
+	}
+	tests := []test{
+		{
+			request: &metrov1.GetSubscriptionRequest{
+				Name: "projects/project123/subscriptions/testsub",
```
vnktram, line 119: Why do we pass name twice? Can we assert `req.Name` with the response rather?

service/web/subscriberserver_test.go
vnktram: @sumitkrzp Keep in mind that while reflections are acceptable in unit tests, we want to avoid using them as part of our actual code. reflections tend to be expensive and should be avoided unless absolutely necessary.

vnktram (changes requested):

service/web/server_test.go
sumitkrzp: @vnktram 
 here request is an interface requesttype will be  different for different resourcename and request may not always contain field name  in test case. Here request.Name means subscriptionName and name means resourceName

sumitkrzp: @vnktram please suggest a way or should i keep it as it is?

service/web/subscriberserver_test.go
sakshi-goyal-razorpay: @sumitkrzp Cannot see the changes.. Also before committing, please verify it is comparing each and every property defined in the object

service/web/server_test.go
```diff
@@ -103,3 +104,32 @@ func TestGetProjectIDFromRequest(t *testing.T) {
...
+	type test struct {
+		request interface{}
+		name    string
+		err     error
+	}
+	type testReq struct {
+	}
+	tests := []test{
+		{
+			request: &metrov1.GetSubscriptionRequest{
```
akashgoyal96, line 118: @sumitkrzp we can also add description about each test case.

sumitkrzp: added comments for the test cases

service/web/server_test.go
sumitkrzp: Added suggested changes for test case

service/web/subscriberserver.go
```diff
@@ -289,3 +289,18 @@ func (s subscriberserver) ListProjectSubscriptions(ctx context.Context,
 func (s subscriberserver) AuthFuncOverride(ctx context.Context, fullMethodName string, req interface{}) (context.Context, error) {
 	return authRequest(ctx, s.credentialCore, fullMethodName, req)
 }
+
+// GetSubscription get subscription details by name
+func (s subscriberserver) GetSubscription(ctx context.Context, req *metrov1.GetSubscriptionRequest) (*metrov1.Subscription, error) {
+	logger.Ctx(ctx).Infow("subscriberserver: received request to get subscription details", "subscription_name", req.GetName())
+	subs, err := s.subscriptionCore.Get(ctx, req.GetName())
+	if err != nil {
+		logger.Ctx(ctx).Errorw("error while getting subscription details from subscription core", "subscription_name", req.GetName())
```
sakshi-goyal-razorpay, line 298: please prepend subscriberserver in this error log message, so that its easy to search for all subscriber server logs

service/web/subscriberserver_test.go
```diff
@@ -684,3 +686,68 @@ func TestSubscriberServer_ListProjectSubscriptionsFailure(t *testing.T) {
...
+	subscriptionCore := mocks2.NewMockICore(ctrl)
+	manager := mocks3.NewMockIManager(ctrl)
+	mockCredentialsCore := mocks5.NewMockICore(ctrl)
+	cache := cachemock.NewMockICache(ctrl)
+	server := newSubscriberServer(mockProjectCore, brokerStore, subscriptionCore, mockCredentialsCore, manager, cache)
+	ctx := context.Background()
+	req := &metrov1.GetSubscriptionRequest{
+		Name: "projects/project123/subscriptions/testsub",
+	}
+	subscriptionCore.EXPECT().Get(gomock.Any(), req.GetName()).Times(1).Return(nil, fmt.Errorf("error"))
```
sakshi-goyal-razorpay, line 749: have a meaningful error message here, although its just a test case, and we are not checking the exact error message, still its more readable that way

service/web/server_test.go
```diff
@@ -103,3 +103,49 @@ func TestGetProjectIDFromRequest(t *testing.T) {
 		assert.Equal(t, tst.err, err)
 	}
 }
+
+func Test_getResourceNameFromRequest(t *testing.T) {
```
sakshi-goyal-razorpay, line 107: I believe this test case should be part of TestGetProjectIDFromRequest function only as that test is internally calling getResourceNameFromRequest(). Also we shouldn't write test case directly for private functions

sakshi-goyal-razorpay (changes requested):

service/web/subscriberserver.go
sumitkrzp: added subscriberserver at the begining

service/web/subscriberserver_test.go
sumitkrzp: added better error message

service/web/server_test.go
sumitkrzp: removed testcase for getResourceNameFromRequest and added in the existing testcase TestGetProjectIDFromReques

sakshi-goyal-razorpay (approved):
LGTM

service/web/subscriberserver_test.go
```diff
@@ -684,3 +686,68 @@ func TestSubscriberServer_ListProjectSubscriptionsFailure(t *testing.T) {
...
+				"test_key": "test_value",
+			},
+			Credentials: &credentials.Model{
+				Username: "test_user",
+				Password: "",
+			},
+		},
+		AckDeadlineSeconds: 10,
+		DeadLetterPolicy: &subscription.DeadLetterPolicy{
+			DeadLetterTopic:     "test_dlq",
```
vnktram, line 719: DLQ should follow our internal naming convention. `{{sub-name}}-dlq`

service/web/subscriberserver_test.go
```diff
@@ -684,3 +686,68 @@ func TestSubscriberServer_ListProjectSubscriptionsFailure(t *testing.T) {
...
+	subscriptionCore.EXPECT().Get(gomock.Any(), req.GetName()).Times(1).Return(sub, nil)
+	res, err := server.GetSubscription(ctx, req)
+	pSub := subscription.ModelToSubscriptionProtoV1(sub)
+	assert.Nil(t, err)
+	assert.NotNil(t, res)
+	assert.True(t, reflect.DeepEqual(res, pSub))
+}
+
+// TestSubscriberServer_GetSubscriptionWithoutExistingName : test the GetSubscription given subscription does not exist
+func TestSubscriberServer_GetSubscriptionWithoutExistingName(t *testing.T) {
```
vnktram, line 736: The package_method format is used to measure coverage. Please do not add the test case identifiers here.

vnktram (changes requested):
please address comments

service/web/subscriberserver_test.go
sumitkrzp: corrected the format of test case

vnktram (approved):

vnktram:
@sumitkrzp LGTM. Let's merge this to master.

PR #402, opened 2022-08-08
Add worker spans as part of publisher trace

sakshi-goyal-razorpay:
For the worker component, the message consumed was creating a new trace rather than using the publisher trace. This PR aims at adding async worker spans to parent publisher span

Screenshot 2022-08-10 at 5 58 52 PM


internal/stream/processor.go
```diff
@@ -155,3 +160,13 @@ func (pr *processor) pushMessage(ctx context.Context, message *metrov1.ReceivedM
...
 	ants.Release()
 }
+
+func (pr *processor) getSpanContext(message *metrov1.ReceivedMessage) opentracing.SpanContext {
+	spanContext, extractErr := opentracing.GlobalTracer().Extract(opentracing.TextMap, opentracing.TextMapCarrier(message.Message.Attributes))
+	if extractErr != nil {
+		logger.Ctx(pr.ctx).Errorw("failed to get span context from message", "error", extractErr.Error())
+		return nil
+	}
+	delete(message.Message.Attributes, messagebroker.UberTraceID)
```
vnktram, line 170: Why are we deleting the traceID from the message? Wouldn't it be re-used at the time of retries?

internal/subscriber/ordering.go
```diff
@@ -132,6 +132,15 @@ func (s *OrderedImplementation) Pull(ctx context.Context, req *PullRequest, resp
 		ts.Seconds = msg.PublishTime.Unix()
 		protoMsg.PublishTime = ts
 
+		if len(protoMsg.Attributes) == 0 {
+			protoMsg.Attributes = make(map[string]string, len(msg.Attributes))
```
vnktram, line 136: If we're only assigning `UberTraceID` to the map, can we avoid initializing one with the size of `msg.Attributes`?

internal/subscriber/subscriberimpl.go
```diff
@@ -97,6 +97,16 @@ func (s *BasicImplementation) Pull(ctx context.Context, req *PullRequest, respon
 		ts.Seconds = msg.PublishTime.Unix()
 		protoMsg.PublishTime = ts
 
+		if len(protoMsg.Attributes) == 0 {
+			protoMsg.Attributes = make(map[string]string, len(msg.Attributes))
```
vnktram, line 101: Please refer above comment.

vnktram (changes requested):

vnktram (approved):

PR #407, opened 2022-08-12
Use Delete to remove entities by key

vnktram:
- Raising a PR to unblock Himanshu.
- Will merge post adding unit tests.

### DIFF
--- a/pkg/registry/consul.go
+++ b/pkg/registry/consul.go
@@ -343,6 +343,25 @@ func (c *ConsulClient) Exists(ctx context.Context, key string) (bool, error) {
 	return true, nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/registry/consul.go
```diff
@@ -343,6 +343,25 @@ func (c *ConsulClient) Exists(ctx context.Context, key string) (bool, error) {
 	return true, nil
 }
 
+// Delete removes the model for a single key form registry
```
sakshi-goyal-razorpay, line 346: single key from*

sakshi-goyal-razorpay (changes requested):

PR #408, opened 2022-08-12
Introduce randomness in the scheduling algorithm even with load balancing

sakshi-goyal-razorpay:
# Description
Today, node bindings refresh end up with the same old bindings because we are assigning subscriptions one by one to the same set of nodes in round robin fashion. To solve for that, going forward, we will
1. Shuffle the subscriptions array before scheduling
2. We will shuffle nodes array before picking up the least loaded node

Fixes https://razorpay.atlassian.net/browse/METRO-112

## Type of change
- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?
- [x] I triggered node binding refresh multiple times to see if subscriptions get assigned to same nodes or not on both local and stage. 
- [x] I checked if the existing functionality of assigning all subscriptions to node acc to the load is intact.

# Is it a breaking change?
No

# Checklist:
- [ ] I have made corresponding changes to the documentation
- [ ] I have added integration tests for the new feature (if applicable)

### DIFF
--- a/internal/scheduler/loadbalancealgoimpl.go
+++ b/internal/scheduler/loadbalancealgoimpl.go
@@ -2,7 +2,9 @@ package scheduler
 
 import (
 	"fmt"
+	"math/rand"
 	"sort"
+	"time"
 
 	"github.com/razorpay/metro/internal/node"
 	"github.com/razorpay/metro/internal/nodebinding"
@@ -48,6 +50,14 @@ func (algo *LoadBalanceAlgoImpl) GetNode(nodebindings []*nodebinding.Model, node
 			Count: v,
 		})
 	}
+
+	rand.Seed(time.Now().UnixNano())
+	rand.Shuffle(
+		len(nodeCountList),
+		func(i, j int) {
+			nodeCountList[i], nodeCountList[j] = nodeCountList[j], nodeCountList[i]
+		},
+	)
 
 	sort.Sort(nodeCountList)
 
--- a/internal/subscription/model.go
+++ b/internal/subscription/model.go
@@ -20,9 +20,6 @@ const (
 	SubscriptionTypePull = "Pull"
 )
 
-// List of subscription model
-type List []*Model
-
 // Model for a subscription
 type Model struct {
 	common.BaseModel
--- a/internal/subscription/model.go
+++ b/internal/subscription/model.go
@@ -20,7 +20,8 @@ const (
 	SubscriptionTypePull = "Pull"
 )
 
-type SubscriptionList []*Model
+// List of subscription model
+type List []*Model
 
 // Model for a subscription
 type Model struct {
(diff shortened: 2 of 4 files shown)


internal/scheduler/loadbalancealgoimpl.go
```diff
@@ -49,6 +51,14 @@ func (algo *LoadBalanceAlgoImpl) GetNode(nodebindings []*nodebinding.Model, node
 		})
 	}
 
+	rand.Seed(time.Now().UnixNano())
```
vnktram, line 54: Let's consider moving this to a util package once we have a fair amount of confidence in generics. Preferably with 1.20.

internal/scheduler/loadbalancealgoimpl.go
```diff
@@ -49,6 +51,14 @@ func (algo *LoadBalanceAlgoImpl) GetNode(nodebindings []*nodebinding.Model, node
...
+	rand.Shuffle(
+		len(nodeCountList),
+		func(i, j int) {
+			nodeCountList[i], nodeCountList[j] = nodeCountList[j], nodeCountList[i]
+		},
+	)
+
 	sort.Sort(nodeCountList)
```
vnktram, line 62: Wouldn't `sort.Sort` nullify the effects of `rand.Shuffle`?

internal/subscription/model.go
```diff
@@ -20,6 +20,9 @@ const (
 	SubscriptionTypePull = "Pull"
 )
 
+// List of subscription model
+type List []*Model
```
vnktram, line 24: Do we need a dedicated type for a collection? Can this be defined as an array where needed?

vnktram (changes requested):

internal/scheduler/loadbalancealgoimpl.go
sakshi-goyal-razorpay: No.. sort will just move the out of order items, but the items having same value will remain as it is

internal/subscription/model.go
sakshi-goyal-razorpay: yes.

sakshi-goyal-razorpay:
> LGTM

Please approve

sumitkrzp (approved):
LGTM

internal/scheduler/loadbalancealgoimpl.go
akashgoyal96: @sakshi-goyal-razorpay since we are using the same piece of code here and in schedulerTask file, can we consider reusing the same code block.

sakshi-goyal-razorpay: I tried that, but without the use of generics, it is not possible, since the array type is different

vnktram (approved):
LGTM. Let's test this on stage.

akashgoyal96 (approved):
LGTM

PR #416, opened 2022-08-29
METRO-125-stream-ut Stream unit test changes

sumitkrzp:
# Description

Added Test cases for stream package.

## Type of change
- [x] Test cases

# How Has This Been Tested?
- [x] By running added test cases

# Checklist:

### DIFF
--- a/internal/stream/manager_test.go
+++ b/internal/stream/manager_test.go
@@ -36,7 +36,7 @@ func TestNewPushStreamManager(t *testing.T) {
 			uuid.New().String(),
 			subName,
 			getSubscriptionCoreMock(ctrl, test.wantErr),
-			getSubscriberCoreMock(t, ctx, ctrl),
+			getSubscriberCoreMock(ctx, t, ctrl),
 			&httpclient.Config{},
 		)
 		assert.Equal(t, test.wantErr, err != nil)
@@ -53,7 +53,7 @@ func TestPushStreamManager_Run(t *testing.T) {
 		uuid.New().String(),
 		subName,
 		getSubscriptionCoreMock(ctrl, false),
-		getSubscriberCoreMock(t, ctx, ctrl),
+		getSubscriberCoreMock(ctx, t, ctrl),
 		&httpclient.Config{},
 	)
 	assert.NoError(t, err)
@@ -78,7 +78,7 @@ func TestPushStreamManager_Stop(t *testing.T) {
 		uuid.New().String(),
 		subName,
 		getSubscriptionCoreMock(ctrl, false),
-		getSubscriberCoreMock(t, ctx, ctrl),
+		getSubscriberCoreMock(ctx, t, ctrl),
 		&httpclient.Config{},
 	)
 	assert.NoError(t, err)
@@ -91,7 +91,7 @@ func TestPushStreamManager_Stop(t *testing.T) {
 	assert.Equal(t, psm.ctx.Err(), context.Canceled)
 }
 
-func getSubscriberCoreMock(t *testing.T, ctx context.Context, ctrl *gomock.Controller) *mocks2.MockICore {
+func getSubscriberCoreMock(ctx context.Context, t *testing.T, ctrl *gomock.Controller) *mocks2.MockICore {
 	subscriberCoreMock := mocks2.NewMockICore(ctrl)
 	subModel := getMockSubModel("")
 
--- a/internal/stream/manager_test.go
+++ b/internal/stream/manager_test.go
@@ -36,7 +36,7 @@ func TestNewPushStreamManager(t *testing.T) {
 			uuid.New().String(),
 			subName,
 			getSubscriptionCoreMock(ctrl, test.wantErr),
-			getSubscriberCoreMock(ctx, ctrl),
+			getSubscriberCoreMock(t, ctx, ctrl),
 			&httpclient.Config{},
 		)
 		assert.Equal(t, test.wantErr, err != nil)
@@ -53,7 +53,7 @@ func TestPushStreamManager_Run(t *testing.T) {
 		uuid.New().String(),
 		subName,
 		getSubscriptionCoreMock(ctrl, false),
-		getSubscriberCoreMock(ctx, ctrl),
+		getSubscriberCoreMock(t, ctx, ctrl),
 		&httpclient.Config{},
 	)
 	assert.NoError(t, err)
@@ -78,7 +78,7 @@ func TestPushStreamManager_Stop(t *testing.T) {
 		uuid.New().String(),
 		subName,
 		getSubscriptionCoreMock(ctrl, false),
-		getSubscriberCoreMock(ctx, ctrl),
+		getSubscriberCoreMock(t, ctx, ctrl),
 		&httpclient.Config{},
 	)
 	assert.NoError(t, err)
@@ -91,7 +91,7 @@ func TestPushStreamManager_Stop(t *testing.T) {
 	assert.Equal(t, psm.ctx.Err(), context.Canceled)
 }
 
-func getSubscriberCoreMock(ctx context.Context, ctrl *gomock.Controller) *mocks2.MockICore {
+func getSubscriberCoreMock(t *testing.T, ctx context.Context, ctrl *gomock.Controller) *mocks2.MockICore {
 	subscriberCoreMock := mocks2.NewMockICore(ctrl)
 	subModel := getMockSubModel("")
 
@@ -104,7 +104,7 @@ func getSubscriberCoreMock(ctx context.Context, ctrl *gomock.Controller) *mocks2
 		defaultMaxOuttandingBytes,
 		gomock.AssignableToTypeOf(make(chan *subscriber.PullRequest)),
 		gomock.AssignableToTypeOf(make(chan *subscriber.AckMessage)),
-		gomock.AssignableToTypeOf(make(chan *subscriber.ModAckMessage))).AnyTimes().Return(getMockSubscriber(ctx, ctrl), nil)
+		gomock.AssignableToTypeOf(make(chan *subscriber.ModAckMessage))).AnyTimes().Return(getMockSubscriber(t, ctx), nil)
 	return subscriberCoreMock
 }
 
--- a/internal/stream/stream_test.go
+++ b/internal/stream/stream_test.go
@@ -260,52 +260,10 @@ func TestNewPushStream_Failure(t *testing.T) {
 	subscriberCoreMock := mocks2.NewMockICore(ctrl)
 	workerID := uuid.New().String()
 	httpConfig := &httpclient.Config{}
-	type args struct {
-		ctx              context.Context
-		nodeID           string
-		subName          string
-		subscriptionCore subscription.ICore
-		subscriberCore   subscriber.ICore
-		config           *httpclient.Config
-	}
-	tests := []struct {
-		name    string
-		args    args
-		want    *PushStream
-		wantErr bool
-	}{
-		{
-			name: "New Push Stream Failure",
-			args: args{
-				ctx:              ctx,
-				nodeID:           workerID,
-				subName:          subName,
-				subscriptionCore: subscriptionCoreMock,
-				subscriberCore:   subscriberCoreMock,
-				config:           httpConfig,
-			},
-			want:    nil,
-			wantErr: true,
-		},
-	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			subscriptionCoreMock.EXPECT().Get(gomock.Any(), subName).DoAndReturn(
-				func(arg0 context.Context, arg1 string) (*subscription.Model, error) {
-					if tt.wantErr {
-						return nil, errors.New("Test Error")
-					}
-					return getMockSubModel(""), nil
-				},
-			)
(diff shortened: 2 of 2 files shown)


internal/stream/stream_test.go
```diff
@@ -250,3 +251,150 @@ func TestPushStream_Stop(t *testing.T) {
...
+				subName:          subName,
+				subscriptionCore: subscriptionCoreMock,
+				subscriberCore:   subscriberCoreMock,
+				config:           httpConfig,
+			},
+			want:    nil,
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
```
sakshi-goyal-razorpay, line 290: if its a single test case, no need to create a test struct array and then loop over tests.

internal/stream/stream_test.go
```diff
@@ -250,3 +251,150 @@ func TestPushStream_Stop(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			subscriptionCoreMock.EXPECT().Get(gomock.Any(), subName).DoAndReturn(
+				func(arg0 context.Context, arg1 string) (*subscription.Model, error) {
+					if tt.wantErr {
+						return nil, errors.New("Test Error")
+					}
+					return getMockSubModel(""), nil
```
sakshi-goyal-razorpay, line 297: this is never executed

internal/stream/stream_test.go
```diff
@@ -250,3 +251,150 @@ func TestPushStream_Stop(t *testing.T) {
...
+				},
+			)
+			got, err := NewPushStream(tt.args.ctx, tt.args.nodeID, tt.args.subName, tt.args.subscriptionCore, tt.args.subscriberCore, tt.args.config)
+			if err == nil {
```
sakshi-goyal-razorpay, line 301: if you are expecting an error, then why this check here..

internal/stream/stream_test.go
```diff
@@ -250,3 +251,150 @@ func TestPushStream_Stop(t *testing.T) {
...
+	}{
+		{
+			name:      "Run Push Stream Restart Subscriber",
+			endpoint:  s.URL,
+			isSuccess: true,
+		},
+	}
+
+	for _, test := range tests {
+		ps, sub := getMockPushStreamRestart(ctx, ctrl, test.endpoint)
```
sakshi-goyal-razorpay, line 333: why defined a separate mock push stream here, and not used getMockPushStream() instead..?

internal/stream/stream_test.go
```diff
@@ -250,3 +251,150 @@ func TestPushStream_Stop(t *testing.T) {
...
+		ps.RunPushStreamManager(ctx)
+		go ps.Start()
+		<-time.NewTicker(10 * time.Millisecond).C
+		ps.subs.GetErrorChannel() <- errors.New("Subscriber Test Error")
+		select {
+		case <-time.NewTicker(1 * time.Second).C:
+			assert.NotEqual(t, ps.subs, sub)
```
sakshi-goyal-razorpay, line 340: you should also check ps.subs is not nil

internal/stream/stream_test.go
```diff
@@ -250,3 +251,150 @@ func TestPushStream_Stop(t *testing.T) {
...
+			ps.stopChan <- true
```
sakshi-goyal-razorpay, line 341: directly stop the stream, as this won't wait for the stream to end

internal/stream/stream_test.go
sumitkrzp: In this test case to restart subscriber I need to multiple changes to the existing one getMockPushStream:
1. I need to return subscriber from the getMockPushStream to compare the old and new subscriber. This needed changes in the existing code.
2. I need to mock getErrorChannel in the getMockSubscriber
3. I need to change the times  of NewSubscriber for restart subscriber
That's why chose to make a new mockPushStream

internal/stream/stream_test.go
sumitkrzp: acknowledged

internal/stream/stream_test.go
sumitkrzp: added check

internal/stream/stream_test.go
sumitkrzp: remove test structs

internal/stream/stream_test.go
sumitkrzp: removed the check

internal/stream/stream_test.go
```diff
@@ -250,3 +251,109 @@ func TestPushStream_Stop(t *testing.T) {
...
+			select {
+			case <-reqCh:
+				if counter < 1 {
+					messages := getMockResponseMessages()
+					resCh <- &metrov1.PullResponse{
+						ReceivedMessages: messages,
+					}
+					counter++
+				}
+				resCh <- &metrov1.PullResponse{}
```
sakshi-goyal-razorpay, line 345: this is a blocking.. resChan is unbuffered channel, and it will block the execution if stream is not listening on this channel. Please wrap this one inside else block.

internal/stream/stream_test.go
sakshi-goyal-razorpay: @sumitkrzp Lets try to reuse the same mock subscriber by making some changes to the existing mock if possible.

internal/stream/stream_test.go
```diff
@@ -250,3 +251,109 @@ func TestPushStream_Stop(t *testing.T) {
...
+			endpoint:  s.URL,
+			isSuccess: true,
+		},
+	}
+
+	for _, test := range tests {
+		ps, sub := getMockPushStreamRestart(ctx, ctrl, test.endpoint)
+		go ps.Start()
+		<-time.NewTicker(10 * time.Millisecond).C
+		if ps.subs != nil {
```
sakshi-goyal-razorpay, line 294: add this as part of assertion

internal/stream/stream_test.go
sakshi-goyal-razorpay: After subscriber is restarted, you should check the new ps.subs is not nil

internal/stream/stream_test.go
sumitkrzp: I tried changing the existing one but i have to make multiple changes since mockSubscriber is written for one time expect and in restart Subscriber we mock to get expect for 2 times, there will be too much if else conditions. So i am using new one only

internal/stream/stream_test.go
sumitkrzp: added in else.

internal/stream/stream_test.go
sumitkrzp: Added

internal/stream/stream_test.go
sumitkrzp: removed

internal/stream/stream_test.go
sakshi-goyal-razorpay: You can just write AnyTimes for the expected func calls. No need to make it condiitonal

sumitkrzp: There was mistake in the test case i was compairing the subscribercoremock with mocksubcriber. fixed the issue. func signature is different now new method is needed.

internal/stream/stream_test.go
```diff
@@ -250,3 +251,90 @@ func TestPushStream_Stop(t *testing.T) {
...
+	got, err := NewPushStream(ctx, workerID, subName, subscriptionCoreMock, subscriberCoreMock, httpConfig)
+	assert.Nil(t, got)
+	assert.NotNil(t, err)
+}
+
+func TestStart_RestartSubscriber(t *testing.T) {
+	ctx := context.Background()
+	ctrl := gomock.NewController(t)
+	ps := getMockPushStreamRestart(t, ctx, ctrl, "")
+	sub := ps.subs
```
sakshi-goyal-razorpay, line 272: move this to line 274

internal/stream/stream_test.go
```diff
@@ -250,3 +251,90 @@ func TestPushStream_Stop(t *testing.T) {
...
+	go ps.Start()
+	<-time.NewTicker(10 * time.Millisecond).C
+	assert.NotNil(t, sub)
+	ps.subs.GetErrorChannel() <- errors.New("Subscriber Test Error")
+	select {
+	case <-time.NewTicker(1 * time.Second).C:
+		assert.NotNil(t, ps.subs)
+		assert.NotEqual(t, sub, ps.subs)
+		ps.Stop()
+	}
```
sakshi-goyal-razorpay, line 282: no need of select case here

internal/stream/stream_test.go
```diff
@@ -250,3 +251,90 @@ func TestPushStream_Stop(t *testing.T) {
...
+}
+
+func getMockPushStreamRestart(t *testing.T, ctx context.Context, ctrl *gomock.Controller, endpoint string) *PushStream {
```
sakshi-goyal-razorpay, line 285: try to reuse getMockPushStream() of stream_test

internal/stream/stream_test.go
```diff
@@ -250,3 +251,90 @@ func TestPushStream_Stop(t *testing.T) {
...
+		defaultTimeoutMs,
+		defaultMaxOutstandingMsgs,
+		defaultMaxOuttandingBytes,
+		gomock.AssignableToTypeOf(make(chan *subscriber.PullRequest)),
+		gomock.AssignableToTypeOf(make(chan *subscriber.AckMessage)),
+		gomock.AssignableToTypeOf(make(chan *subscriber.ModAckMessage))).Return(getMockSubscriberRestart(t, ctx, ctrl), nil).Times(2)
+	return pushStream
+}
+
+func getMockSubscriberRestart(t *testing.T, ctx context.Context, ctrl *gomock.Controller) *mocks3.MockISubscriber {
```
sakshi-goyal-razorpay, line 307: no need to pass ctrl here

internal/stream/stream_test.go
```diff
@@ -250,3 +251,90 @@ func TestPushStream_Stop(t *testing.T) {
...
+	subscriberMock := mocks3.NewMockISubscriber(gomock.NewController(t))
```
sakshi-goyal-razorpay, line 308: do add a comment on why we need to initialize with a new controller here

internal/stream/stream_test.go
sumitkrzp: Not able use older one, because if mockSubscriber passed as the parameter of older method, newsubscriber expect call will always give same instance of mockSubscriber. And to get new Instance  we must use **DoAndReturn** instead of **Return**.

internal/stream/stream_test.go
sumitkrzp: Moved

internal/stream/stream_test.go
sumitkrzp: removed select

internal/stream/stream_test.go
sumitkrzp: removed from parameter.

internal/stream/stream_test.go
sumitkrzp: Added

internal/stream/stream_test.go
```diff
@@ -251,3 +252,93 @@ func TestPushStream_Stop(t *testing.T) {
...
+		workerID,
+		subModel,
+		defaultTimeoutMs,
+		defaultMaxOutstandingMsgs,
+		defaultMaxOuttandingBytes,
+		gomock.AssignableToTypeOf(make(chan *subscriber.PullRequest)),
+		gomock.AssignableToTypeOf(make(chan *subscriber.AckMessage)),
+		gomock.AssignableToTypeOf(make(chan *subscriber.ModAckMessage))).DoAndReturn((func(arg0 context.Context, arg1 string, arg2 *subscription.Model,
+		arg3 int, arg4, arg5 int64, arg6 chan *subscriber.PullRequest, arg7 chan *subscriber.AckMessage, arg8 chan *subscriber.ModAckMessage) (subscriber.ISubscriber, error) {
+		return getMockSubscriberRestart(t, ctx), nil
```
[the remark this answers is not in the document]
sumitkrzp: Since we need new Instance of subscriber everytime newSubscriber is called, Moved this code to older one

internal/stream/stream_test.go
sumitkrzp: Moved this code to older one

internal/stream/manager_test.go
```diff
@@ -104,7 +104,7 @@ func getSubscriberCoreMock(ctx context.Context, ctrl *gomock.Controller) *mocks2
 		defaultMaxOuttandingBytes,
 		gomock.AssignableToTypeOf(make(chan *subscriber.PullRequest)),
 		gomock.AssignableToTypeOf(make(chan *subscriber.AckMessage)),
-		gomock.AssignableToTypeOf(make(chan *subscriber.ModAckMessage))).AnyTimes().Return(getMockSubscriber(ctx, ctrl), nil)
+		gomock.AssignableToTypeOf(make(chan *subscriber.ModAckMessage))).AnyTimes().Return(getMockSubscriber(t, ctx), nil)
```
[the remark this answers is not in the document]
sumitkrzp: golang cli linting is not checking all the files

sakshi-goyal-razorpay (approved):

PR #424, opened 2022-09-20
Increasing UT coverage

sakshi-goyal-razorpay:
# Description
Added unit tests to increase unit test coverage

Fixes https://razorpay.atlassian.net/browse/METRO-158

## Type of change
- [x] Increasing test coverage

# How Has This Been Tested?
- [x] Running unit tests on local

# Is it a breaking change?
No

# Checklist:
- [ ] I have made corresponding changes to the documentation
- [ ] I have added integration tests for the new feature (if applicable)

### DIFF
--- a/internal/credentials/repo_test.go
+++ b/internal/credentials/repo_test.go
@@ -1,0 +1,45 @@
+package credentials
+
+import (
+	"context"
+	"encoding/json"
+	"reflect"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/razorpay/metro/internal/common"
+	"github.com/razorpay/metro/pkg/registry"
+	"github.com/razorpay/metro/pkg/registry/mocks"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRepo_List(t *testing.T) {
+	ctx := context.Background()
+	ctrl := gomock.NewController(t)
+	mockRegistry := mocks.NewMockIRegistry(ctrl)
+	r := NewRepo(mockRegistry)
+	dummyModel := getDummyModel()
+
+	mockRegistry.EXPECT().List(gomock.Any(), "prefix").Return(getDummyRegistryPairs(dummyModel), nil)
+	got, err := r.List(ctx, "prefix")
+	assert.NoError(t, err)
+	if !reflect.DeepEqual(got, []common.IModel{dummyModel}) {
+		t.Errorf("List got %v, want %v", got, []common.IModel{dummyModel})
+	}
+}
+
+func getDummyModel() *Model {
+	model := &Model{Username: "username", Password: "password", ProjectID: "project_id"}
+	model.SetVersion("version")
+	return model
+}
+
+func getDummyRegistryPairs(model common.IModel) []registry.Pair {
+	value, _ := json.Marshal(model)
+	return []registry.Pair{
+		{
+			Value:   []byte(value),
+			Version: model.GetVersion(),
+		},
+	}
+}
--- a/internal/credentials/repo_test.go
+++ b/internal/credentials/repo_test.go
@@ -23,9 +23,7 @@ func TestRepo_List(t *testing.T) {
 	mockRegistry.EXPECT().List(gomock.Any(), "prefix").Return(getDummyRegistryPairs(dummyModel), nil)
 	got, err := r.List(ctx, "prefix")
 	assert.NoError(t, err)
-	if !reflect.DeepEqual(got, []common.IModel{dummyModel}) {
-		t.Errorf("List got %v, want %v", got, []common.IModel{dummyModel})
-	}
+	assert.True(t, reflect.DeepEqual(got, []common.IModel{dummyModel}))
 }
 
 func getDummyModel() *Model {
--- a/internal/credentials/repo_test.go
+++ b/internal/credentials/repo_test.go
@@ -38,7 +38,7 @@ func getDummyRegistryPairs(model common.IModel) []registry.Pair {
 	value, _ := json.Marshal(model)
 	return []registry.Pair{
 		{
-			Value:   []byte(value),
+			Value:   value,
 			Version: model.GetVersion(),
 		},
 	}
--- a/internal/project/core_test.go
+++ b/internal/project/core_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
+	"github.com/razorpay/metro/internal/common"
 	mocks "github.com/razorpay/metro/internal/project/mocks/repo"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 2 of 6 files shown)


sakshi-goyal-razorpay:
This is the first PR that attempts to increase the code coverage by 1.85%. I will create one more PR to increase it further to 65%

internal/credentials/repo_test.go
```diff
@@ -0,0 +1,45 @@
...
+	ctx := context.Background()
+	ctrl := gomock.NewController(t)
+	mockRegistry := mocks.NewMockIRegistry(ctrl)
+	r := NewRepo(mockRegistry)
+	dummyModel := getDummyModel()
+
+	mockRegistry.EXPECT().List(gomock.Any(), "prefix").Return(getDummyRegistryPairs(dummyModel), nil)
+	got, err := r.List(ctx, "prefix")
+	assert.NoError(t, err)
+	if !reflect.DeepEqual(got, []common.IModel{dummyModel}) {
```
sumitkrzp, line 26: Here we can asser.true(t, DeepEqual(got, []common.IModel{dummyModel}))

internal/project/core_test.go
```diff
@@ -230,3 +231,34 @@ func TestCore_DeleteProject(t *testing.T) {
...
+	}{
+		{
+			expected: []string{"project_1", "project_2"},
+			wantErr:  false,
+			err:      nil,
+		},
+		{
+			expected: nil,
+			wantErr:  true,
+			err:      fmt.Errorf("Something went wrong"),
```
sumitkrzp, line 254: Is this err required because we are not asserting any particular error?

internal/publisher/core_test.go
```diff
@@ -0,0 +1,74 @@
...
+			expected: []string{msgID},
+			wantErr:  false,
+			err:      nil,
+		},
+		{
+			req:      req,
+			msgID:    "",
+			expected: nil,
+			wantErr:  true,
+			err:      fmt.Errorf("Something went wrong"),
```
sumitkrzp, line 45: *SAA

internal/publisher/core_test.go
```diff
@@ -0,0 +1,74 @@
...
+func TestCore_Publish(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	ctx := context.Background()
+	mockBrokerStore := brokerstoremock.NewMockIBrokerStore(ctrl)
+	producer := mockMB.NewMockProducer(ctrl)
+	publisherCore := NewCore(mockBrokerStore)
+	req := getDummyPublishRequest()
+	msgID := xid.New().String()
+
+	tests := []struct {
```
sumitkrzp, line 26: you include name of test in test struct  and run test case like `for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {}}`
this will log test name if test case is failing

internal/subscriber/sequence_test.go
```diff
@@ -192,20 +192,53 @@ func Test_offsetSequenceManager_GetOrderedSequenceNum(t *testing.T) {
...
-		})
-
 	offsetSeqManager := getMockSequenceManager(ctx, mockRepo)
 	dummyMsg := getDummyReceivedMessage()
-	got, err := offsetSeqManager.GetOrderedSequenceNum(ctx, getMockSubModel(), dummyMsg)
-	assert.NoError(t, err)
-	assert.Equal(t, dummyMsg.Offset, got.CurrentSequenceNum)
-	assert.Equal(t, previousOffset, got.PrevSequenceNum)
+
+	tests := []struct {
```
sumitkrzp, line 198: SAA

internal/project/core_test.go
sakshi-goyal-razorpay: if it is not added as part of test struct, then we will need to add conditional statements which is not ideal

internal/publisher/core_test.go
sakshi-goyal-razorpay: SAA

internal/publisher/core_test.go
sakshi-goyal-razorpay: test names are logged even without the above statement. You can check it on github as well when the unit tests are failing as well as on local

internal/subscriber/sequence_test.go
sakshi-goyal-razorpay: SAA

internal/publisher/core_test.go
sumitkrzp: the test with multiple test we will not be able to get which of those cases are missing

internal/subscriber/sequence_test.go
sumitkrzp: SAA

internal/publisher/core_test.go
sakshi-goyal-razorpay: Screenshot 2022-09-14 at 4 18 42 PM

Multiple tests case are failing. Do you have some doc that says otherwise..?

sumitkrzp: If test have multiple test cases we will be able to identify which case is failing

internal/project/core_test.go
sumitkrzp: wantErr is being used for that only , we can assert.Equal(wantErr, err!=nil)

sakshi-goyal-razorpay: The problem is with line no 263. I need to mock the error response. I don't want to have conditional statement on wantErr to return the error response

sumitkrzp (approved):
LGTM

internal/publisher/core_test.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+func TestCore_Publish(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	ctx := context.Background()
+	mockBrokerStore := brokerstoremock.NewMockIBrokerStore(ctrl)
+	producer := mockMB.NewMockProducer(ctrl)
+	publisherCore := NewCore(mockBrokerStore)
+	req := getDummyPublishRequest()
+	msgID := xid.New().String()
```
akashgoyal96, line 24: @sakshi-goyal-razorpay why are we using xid instead of UUID here?

internal/publisher/core_test.go
```diff
@@ -0,0 +1,82 @@
...
+		req      *metrov1.PublishRequest
+		msgID    string
+		expected []string
+		wantErr  bool
+		err      error
+	}{
+		{
+			name:     "Message publishing with no errors",
+			req:      req,
+			msgID:    msgID,
```
akashgoyal96, line 37:
```suggestion
			msgID:    xid.New().String(),
```

akashgoyal96 (changes requested):
Added a couple of comments, otherwise looks fine

internal/publisher/core_test.go
sakshi-goyal-razorpay: This is what we use while assigning message id
https://github.com/razorpay/metro/blob/master/pkg/messagebroker/kafka_utils.go#L220

internal/publisher/core_test.go
sakshi-goyal-razorpay: can't do, need to use same value of msgID in msgID and expected field

akashgoyal96 (approved):
LGTM

PR #426, opened 2022-09-26
METRO-143 Cleanup topics from kafka on subscription/topic deletion

sumitkrzp:
# Description
Clean-up of kafka topics when delete topic or subscription is being deleted

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?
Tested on local by making delete calls and checked on kafdrop if the topics are deleted in kafka or not.

# Is it a breaking change?
Yes, this change cannot be test via canary, since delete calls are very less to compare.

# Checklist:
- [x] I have added unit tests that prove my fix is effective or that my feature works


service/web/subscriberserver.go
```diff
@@ -203,7 +204,11 @@ func (s subscriberserver) DeleteSubscription(ctx context.Context, req *metrov1.D
 	if err != nil {
 		return nil, merror.ToGRPCError(err)
 	}
-
+	sub, err := s.subscriptionCore.Get(ctx, req.Subscription)
```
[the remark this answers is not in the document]
sumitkrzp: added check for err

sakshi-goyal-razorpay:
@sumitkrzp Request review only when all the PR checks are working,

internal/brokerstore/brokerstore_test.go
```diff
@@ -117,3 +118,41 @@ func getValidProducerClientOptions() messagebroker.ProducerClientOptions {
...
+	type args struct {
+		ctx context.Context
+	}
+	ctx := context.Background()
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+		want   bool
+	}{
```
subintp, line 136: add -ve test case as well

internal/subscription/core.go
```diff
@@ -497,3 +501,60 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
 	logger.Ctx(ctx).Infow("migration: request completed.", "subscriptionsUpdated", updatedSubCount)
 	return nil
 }
+
+func (c *Core) deleteSubscriptionTopics(ctx context.Context, m *Model) error {
+	if m == nil || m.GetTopic() == "" {
```
subintp, line 506: Why do we have to write all delete logic here? Can we separate it into a different struct or helper?

internal/subscription/core_test.go
```diff
@@ -307,12 +308,107 @@ func TestCore_DeleteSubscription(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockProjectCore.EXPECT().ExistsWithID(gomock.Any(), tt.args.subscription.ExtractedSubscriptionProjectID).DoAndReturn(func(arg0 context.Context, arg1 string) (bool, error) {
+				if arg1 == tt.args.subscription.ExtractedSubscriptionProjectID {
+					return true, nil
+				} else {
+					return false, errors.New("project not found")
+				}
+			}).AnyTimes()
```
subintp, line 396: These tests doesn't cover the correctness end to end as we as we are checking only the err retired. 

Please add test cases to assert `deleteSubscriptionTopic`  is called the with **right topic name** for subscription, retry, delay and dlq

internal/topic/core_test.go
```diff
@@ -547,3 +550,178 @@ func TestCore_Get(t *testing.T) {
...
+			projectCore.EXPECT().ExistsWithID(gomock.Any(), tt.args.m.ExtractedProjectID).DoAndReturn(func(arg0 context.Context, arg1 string) (bool, error) {
+				if arg1 == extracteProjectID {
+					return true, nil
+				} else {
+					return false, errors.New("ProjectID not found")
+				}
+			}).AnyTimes()
+			projectCore.EXPECT().Exists(tt.args.ctx, gomock.Any()).Return(true, nil).AnyTimes()
+			brokerStore.EXPECT().IsTopicCleanUpEnabled(gomock.Any()).Return(true)
+			repo.EXPECT().Delete(gomock.Any(), tt.args.m).Return(nil).AnyTimes()
```
subintp, line 628: Assert DeleteTopic with the right topic name in the tests.

subintp (changes requested):

internal/subscription/core_test.go
sumitkrzp: @subintp right topic name test case is already ```TestCore_DeleteSubscription``` written, i this test case i am only checking if we get any error while deleting any subscription related topics

internal/topic/core_test.go
sumitkrzp: @subin already written test cases for the right topic name please check line no. 586

internal/subscription/core.go
sumitkrzp: @subin this is helper function only, we cannot separate this logic anyway as we need to delete all subscription related topic here i am passing only name of different topic and logic to delete from broker is written on different method only.

internal/subscription/core_test.go
subintp: 1. `TestCore_DeleteSubscription` tests what is happening inside that method. 
2.  `deleteSubscriptionTopic` is called multiple times from `deleteSubscriptionTopics` with multiple topic names. The correctness of `deleteSubscriptionTopics`  depends on whether we are passing the right topic name to `deleteSubscriptionTopic` method which is not covered in the test cases.

subintp (changes requested):

internal/brokerstore/brokerstore.go
```diff
@@ -325,3 +328,9 @@ func (b *BrokerStore) FlushAllProducers(ctx context.Context) {
 	})
 	wg.Wait()
 }
+
+// IsTopicCleanUpEnabled checks if topic clean-up config is enabled or not
+func (b *BrokerStore) IsTopicCleanUpEnabled(ctx context.Context) bool {
+	logger.Ctx(ctx).Infof("EnableTopicCleanUp value: %v", b.bConfig.Admin.EnableTopicCleanUp)
```
sakshi-goyal-razorpay, line 334: do we really need this log..?

internal/subscription/core.go
```diff
@@ -497,3 +501,60 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
...
+		return errors.New("Nil model")
```
sakshi-goyal-razorpay, line 507: make it verbose

internal/subscription/core.go
```diff
@@ -497,3 +501,60 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
...
+	}
+
+	// delete delay topics
+	for _, delayTopic := range m.GetDelayTopics() {
+		delayModel := &topic.Model{
+			Name:               delayTopic,
+			ExtractedProjectID: projectID,
+			ExtractedTopicName: topic.GetTopicNameOnly(delayTopic),
+		}
+		err = c.topicCore.DeleteSubscriptionTopic(ctx, delayModel)
```
sakshi-goyal-razorpay, line 533: why don't we use c.deleteSubscriptionTopic here..?

internal/subscription/model.go
```diff
@@ -19,6 +20,9 @@ const (
 
 	// SubscriptionTypePull ...
 	SubscriptionTypePull = "Pull"
+
+	// DeadLetterTopicSuffix dead letter topic suffix
+	DeadLetterTopicSuffix = "-dlq"
```
sakshi-goyal-razorpay, line 25: why not reuse DeadLetterTopicSuffix defined in internal/topic package..?

internal/topic/core.go
```diff
@@ -185,6 +186,47 @@ func (c *Core) DeleteTopic(ctx context.Context, m *Model) error {
...
+
+	if ok, err := c.projectCore.ExistsWithID(ctx, m.ExtractedProjectID); !ok {
+		if err != nil {
+			return err
+		}
+		return merror.Newf(merror.NotFound, "Project not found")
+	}
+	// clean-up topic from broker
+	if err := c.DeleteBrokerTopic(ctx, m); err != nil {
+		return err
```
sakshi-goyal-razorpay, line 216: if cleanup is not enabled, this will pass as deleteBrokerTopic will return nil, but data will be deleted from consul for non retry/internal topics. Is that intentional..?

internal/subscription/core.go
```diff
@@ -497,3 +501,60 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
...
+
+func (c *Core) deleteSubscriptionTopics(ctx context.Context, m *Model) error {
+	if m == nil || m.GetTopic() == "" {
+		return errors.New("Nil model")
+	}
+
+	projectID := m.ExtractedTopicProjectID
+
+	//  delete internal topic
+	internalTopic := m.GetSubscriptionTopic()
```
vnktram, line 513: build an array of topics to be deleted and execute in a loop. This is cumbersome to maintain

internal/subscription/core.go
```diff
@@ -497,3 +501,60 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
...
+	err = c.deleteSubscriptionTopic(ctx, dlqTopic, projectID, topic.GetTopicNameOnly(dlqTopic))
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *Core) deleteSubscriptionTopic(ctx context.Context, name, projectID, extractedTopicName string) error {
+	return c.topicCore.DeleteSubscriptionTopic(ctx, &topic.Model{
```
vnktram, line 555: Can this be made a bulk operation rather than multiple invocations?

internal/subscription/core_test.go
vnktram: +1. We never test which topics get deleted in the first place. This is the core functionality that needs to get tested. How do we know if we are deleting all relevant topics here?

vnktram (changes requested):

internal/topic/core.go
sumitkrzp: @sakshi-goyal-razorpay  yes this is intentional as config will tell if we want to delete kafka topics or not

internal/brokerstore/brokerstore_test.go
sumitkrzp: added

internal/subscription/core.go
sumitkrzp: changed in the new commit

internal/subscription/core_test.go
sumitkrzp: Added test case to test right names

internal/brokerstore/brokerstore.go
sumitkrzp: removed

internal/subscription/core.go
sumitkrzp: added

internal/subscription/core.go
sumitkrzp: changes in the method

internal/subscription/model.go
sumitkrzp: used topic suffix nmow

internal/subscription/core.go
sumitkrzp: Done

internal/subscription/core.go
sumitkrzp: deleted  the mathod, made changes in deleteSubscription topics

service/web/subscriberserver.go
```diff
@@ -203,7 +204,14 @@ func (s subscriberserver) DeleteSubscription(ctx context.Context, req *metrov1.D
...
+	if err != nil {
+		return &emptypb.Empty{}, merror.ToGRPCError(merror.Newf(merror.NotFound, "Subscription does not exist"))
+	}
+	m.Topic = sub.Topic
+	m.ExtractedSubscriptionName = sub.ExtractedSubscriptionName
+	m.ExtractedSubscriptionProjectID = sub.ExtractedSubscriptionProjectID
```
sakshi-goyal-razorpay, line 213: we are setting this inside GetValidatedModelForDelete func, right..?

config/perf.toml
```diff
@@ -23,6 +23,8 @@
             brokers       = ["localhost:9092"]
             enableTLS     = true
             certDir       = "/app/configs/"
+            [web.broker.brokerconfig.admin]
+                enableTopicCleanUp = false
```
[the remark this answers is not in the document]
vnktram: @sakshi-goyal-razorpay Please address.

internal/subscription/core.go
```diff
@@ -497,3 +501,47 @@ func (c *Core) Migrate(ctx context.Context, names []string) error {
...
+			}
+		} else if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// getSubsTopics gets list of all the subscription topics
+func getSubsTopics(m *Model) []string {
```
vnktram, line 537:
```suggestion
func getSubcriptionTopics(m *Model) []string {
```

vnktram (changes requested):

internal/subscription/core.go
vnktram, line 537:
```suggestion
func getSubscriptionTopics(m *Model) []string {
```

vnktram (approved):

subintp (approved):

PR #439, opened 2022-10-07
Revert "Revert "Watcher implementation as publisher task""

akashgoyal96:
Reverts razorpay/metro#435

### DIFF
--- a/internal/tasks/publishertask.go
+++ b/internal/tasks/publishertask.go
@@ -1,0 +1,163 @@
+package tasks
+
+import (
+	"context"
+
+	"github.com/razorpay/metro/internal/common"
+	"github.com/razorpay/metro/internal/topic"
+	"github.com/razorpay/metro/pkg/logger"
+	"github.com/razorpay/metro/pkg/registry"
+
+	"golang.org/x/sync/errgroup"
+)
+
+// PublisherTask implements the Watcher and maintains a pre-warmup.
+type PublisherTask struct {
+	id             string
+	registry       registry.IRegistry
+	topicCore      topic.ICore
+	topicWatchData chan *struct{}
+}
+
+// topicCacheData is declared Global to keep it instance agnostic
+// NOTE: Only read queries to be written from Public methods on topicCacheData
+var topicCacheData map[string]bool = make(map[string]bool)
+
+// NewPublisherTask creates PublisherTask instance
+func NewPublisherTask(
+	id string,
+	registry registry.IRegistry,
+	topicCore topic.ICore,
+	options ...Option,
+) (ITask, error) {
+
+	publisherTask := &PublisherTask{
+		id:             id,
+		registry:       registry,
+		topicCore:      topicCore,
+		topicWatchData: make(chan *struct{}),
+	}
+
+	for _, option := range options {
+		option(publisherTask)
+	}
+
+	return publisherTask, nil
+}
+
+// Run the task
+func (pu *PublisherTask) Run(ctx context.Context) error {
+	var err error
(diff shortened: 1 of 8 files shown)


sakshi-goyal-razorpay (approved):

service/web/publisherserver.go
```diff
@@ -29,16 +30,38 @@ func newPublisherServer(projectCore project.ICore, brokerStore brokerstore.IBrok
...
+		topicCacheCount.WithLabelValues(env, "Hit", req.Topic).Inc()
+	} else {
+		topicCacheCount.WithLabelValues(env, "Miss", req.Topic).Inc()
+		logger.Ctx(ctx).Infow(
+			"PublishServer: Topic doesn't exist inside the cache..",
+			"req",
+			req.Topic)
+
+		// Fallback method to check topic existence in the Registry
+		if ok, err := s.topicCore.ExistsWithName(
```
vnktram, line 54: If topic is not found in cache but is present in the registry then we should update the local cache data with the topic info.

internal/tasks/publishertask.go
```diff
@@ -0,0 +1,170 @@
...
+
+		return terr
+	}
+
+	topicData := make(map[string]bool)
+	for _, topic := range topics {
+		topicData[topic.Name] = true
+	}
+
+	TopicCacheData = topicData
```
vnktram, line 150: Wrap this around a semaphore.

internal/tasks/publishertask.go
```diff
@@ -0,0 +1,170 @@
...
+	if _, ok := TopicCacheData[topic]; ok {
+
+		return true
+	}
+
+	return false
+}
+
+// UpdateTopicCache is to update the cache
+func UpdateTopicCache(ctx context.Context, topic string) {
```
sakshi-goyal-razorpay, line 167: you can have a flag in this function that defines whether entire cache needs to be updated or just for topic. 
and based on that, you can update the topicCacheData inside a semaphore.
You can define a struct here, to handle both cases.

internal/tasks/publishertask.go
```diff
@@ -0,0 +1,183 @@
...
+
+		return true
+	}
+
+	return false
+}
+
+// UpdateTopicCache is to update the cache
+func UpdateTopicCache(ctx context.Context, topicMap map[string]bool, specificTopicBool bool) {
+	sem <- 1
```
vnktram, line 170: Let's use a simple Mutex lock and avoid re-inventing the wheel here. https://pkg.go.dev/sync#Mutex.Lock

internal/tasks/publishertask.go
```diff
@@ -0,0 +1,183 @@
...
+	if _, ok := TopicCacheData[topic]; ok {
+
+		return true
+	}
+
+	return false
+}
+
+// UpdateTopicCache is to update the cache
+func UpdateTopicCache(ctx context.Context, topicMap map[string]bool, specificTopicBool bool) {
```
vnktram, line 169: `specificTopicBool` needs better documentation. If possible avoid passing a bool altogether and find a better way to update the cache.

vnktram (changes requested):

internal/tasks/publishertask.go
akashgoyal96: I have added detailed comment for `specificTopicBool `

internal/tasks/publishertask.go
vnktram: Why was this marked as resolved? Can we use a mutex instead?

vnktram (approved):

PR #453, opened 2022-11-15
Added new metric to measure the latency for push endpoint

sakshi-goyal-razorpay:
# Description
Added new metric to measure the latency for push endpoint

## Type of change
- [ ] Added new metric to figure out latency difference

### DIFF
--- a/internal/stream/metrics.go
+++ b/internal/stream/metrics.go
@@ -16,6 +16,7 @@ var (
 	workerSubscriberErrors           *prometheus.CounterVec
 	workerPushEndpointCallsCount     *prometheus.CounterVec
 	workerComponentRestartCount      *prometheus.CounterVec
+	workerPushEndpointLatency        *prometheus.HistogramVec
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


internal/stream/metrics.go
```diff
@@ -50,4 +51,10 @@ func init() {
 	workerComponentRestartCount = promauto.NewCounterVec(prometheus.CounterOpts{
 		Name: "metro_worker_component_restart_count",
 	}, []string{"env", "component", "topic", "subscription"})
+
+	workerPushEndpointLatency = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "metro_worker_push_endpoint_latency_seconds",
+		Help:    "Time taken to get response from push endpoint",
+		Buckets: prometheus.ExponentialBuckets(0.001, 1.25, 300),
```
mehuled, line 58: could not understand the reasoning behind these buckets? why spaced at 1.25 and 1.75?

mehuled (approved):

internal/stream/metrics.go
sakshi-goyal-razorpay: Bucket spacing is used same as push endpoint metric.

akashgoyal96 (approved):
LGTM

